Issue #67, opened 2026-09-01
External tool versions drift silently between machines; nothing reports them


jerimiah797:
## Summary

Quern installs six external tools and pins none of them. Every machine ends up with whatever was current the day `quern setup` ran, and there is no way to see the difference — `check_tools()` reports availability as booleans, never versions.

This is not theoretical. Two machines working on the same investigation are three majors apart on `pymobiledevice3`:

| Machine | `pymobiledevice3` |
|---|---|
| work machine | 7.7.1 |
| `scorpius` | 9.15.1 |
| PyPI current | 11.3.0 |

Neither machine did anything wrong. It surfaced as one machine reporting that `webinspector` was lockdown/RSD-only while the other read different code — a "works on mine" that cost real time to resolve, and would have been a one-command answer if either could see the other's versions.

## Why not pinning

Pinning six external tools means owning six upgrade treadmills. For a solo project that is likely worse than the drift.

It also would not work. Tools move *after* install: `brew upgrade` can shift `mitmproxy`, `libimobiledevice` or `node` underneath Quern at any time, and every user's upgrade cadence is different and unknowable. Quern cannot control when these change, so the goal should be **detect and report**, not constrain.

`python@3.12` is the only version constraint anywhere in `setup.py` today.

## Proposal: "tested with", reported at runtime

Four stages, each useful alone. Stage 1 delivers most of the value.

### Stage 1 — report versions

Extend `check_tools()` (`server/device/controller.py:75`) to return a version alongside availability, and show it in `quern doctor` and `GET /tools`.

Useful entirely on its own: no policy, no manifest, nothing to maintain, and it is what turns "works on mine, not yours" into one command. Everything below depends on it — you cannot compare what you cannot see.

**Work:** per-tool version parsing. `_get_version()` in `setup.py:200` takes the first non-empty line, which is not enough — the formats disagree:

```
pymobiledevice3  →  9.15.1
adb              →  Android Debug Bridge version 1.0.41
idevice_id       →  idevice_id 1.4.0
node             →  v22.22.2
```

Also needs to resolve tools the way Quern already does rather than relying on `PATH` — `idb` and `mitmdump` are not on a plain shell's `PATH` on at least one machine.

### Stage 2 — declare what was tested

A `tested_with` manifest in the repo. **Not pins, not floors** — a record of the versions a release was actually exercised against.

`quern doctor` compares and reports below/above, without judgement about whether either is wrong. Being above the tested version is useful information, not an error.

[...]

PR #70, opened 2026-09-01
Simulator Web Inspector transport for webview DOM access

jerimiah797:
Lands the transport for reading DOM inside `WKWebView`s on simulators. **Discovery and enumeration only** — evaluation and interaction are deliberately left out so the protocol layer can be reviewed on its own.

## Why

The accessibility tree does not descend into `WKWebView` on simulators. A screen showing a fully rendered page returns five native-chrome elements and nothing else — measured on Geocaching's Shareables screen, where XCUITest sees 47 elements and WDA 63 for the same screen at the same moment.

WebKit does expose that content, via the Web Inspector protocol. On a simulator it is a plain unix socket under `/private/tmp` — no usbmux, no lockdown, no pairing, no developer disk image.

## Approach

`pymobiledevice3.ServiceConnection` already speaks the wire format (length-prefixed plists carrying `__selector` / `__argument` pairs) and its constructor accepts a raw socket. So this supplies the transport and drives it directly, rather than going through `WebinspectorService`, which requires a lockdown object a simulator does not have.

Deliberately **not** a fork, a monkey-patch, or a vendored copy. Nothing in `pymobiledevice3` needs changing.

## Three bugs found by running it, not reasoning about it

All three would have shipped otherwise.

**Socket files are never cleaned up.** When a simulator shuts down its socket file remains. This machine had **16 candidate paths and 1 live listener**, so selecting the first match by name reliably connects to a corpse. Every candidate is now probed, newest first, and the error reports how many were tried rather than claiming no simulator is booted.

**Applications arrive in waves.** The first `_rpc_reportConnectedApplicationList:` is frequently partial, with more following as `_rpc_applicationConnected:`. Stopping at the first list found **1 application where there were 3** — and the one it missed was the one showing content. The handshake now drains until the channel goes quiet.

**The dependency import was ordered before socket discovery** — caught by the pre-commit suite, not by my own tests. A missing dependency masked the more useful "no simulator is booted". Socket probing needs only the standard library, so the import now happens once a live socket is in hand.

## The dependency floor is load-bearing

`pymobiledevice3>=11.0`, and this matters more than a typical floor:

| | 7.7.1 | 9.15.1 | 11.3.0 |
|---|---|---|---|
| `send_plist` / `recv_plist` | sync | sync | **coroutines** |
| `aio_send_plist` / `aio_recv_plist` | present | present | **removed** |

The same method names changed semantics. Below the floor the call returns an un-awaited coroutine and **does nothing at all** — no exception, just a `RuntimeWarning`. Verified against all three versions.

This also makes `pymobiledevice3` a **declared Python dependency**, where today it is installed by setup and consumed only as a CLI. That is a change worth a look, though arguably an improvement for #67: a declared dependency can be floored, whereas the CLI install drifts silently and invisibly.

## Testing

Three unit tests cover socket selection, including that a dead socket file is not mistaken for an absent simulator. Full suite passes (1443).

The protocol paths need a booted simulator and are not unit-testable here. They were exercised manually against iOS 18.6 — enumerating applications and pages, and cross-checked against raw protocol output to confirm the parse rather than trust it.

## Not included

`Runtime.evaluate`, page interaction, MCP tool surface. Those need `Target.sendMessageToTarget` wrapping — flat `Runtime.evaluate` fails with `'Runtime' domain was not found` — which belongs with the evaluation layer rather than the transport.

Background and the decision trail: `docs/proposals/webview-channel-log.md` entries 5, 13, 14 and `webview-a11y-spike-findings.md`.

## Summary by CodeRabbit

* **New Features**
  * Added support for inspecting iOS Simulator web content through the Web Inspector.
  * Applications and their inspectable pages can now be discovered through the simulator.
  * Improved handling of pending page listings and unavailable or stale inspection connections.

* **Bug Fixes**
  * Added clearer feedback when simulator inspection channels cannot be reached.

* **Tests**
  * Added coverage for simulator socket discovery, connection ordering, and stale connection handling.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,6 +8,11 @@ requires-python = ">=3.11"
 dependencies = [
     "fastapi>=0.115.0",
     "uvicorn[standard]>=0.34.0",
+    # Web Inspector transport for simulator webview inspection. Floored at 11.0
+    # deliberately: send_plist/recv_plist are synchronous in <=9.x and coroutines
+    # from 11.x, so an older release fails silently -- the call returns an
+    # un-awaited coroutine and does nothing. See server/device/webinspector.py.
+    "pymobiledevice3>=11.0",
     "pydantic>=2.0",
     "sse-starlette>=2.0",
     "mitmproxy>=10.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,11 +8,13 @@ requires-python = ">=3.11"
 dependencies = [
     "fastapi>=0.115.0",
     "uvicorn[standard]>=0.34.0",
-    # Web Inspector transport for simulator webview inspection. Floored at 11.0
-    # deliberately: send_plist/recv_plist are synchronous in <=9.x and coroutines
-    # from 11.x, so an older release fails silently -- the call returns an
-    # un-awaited coroutine and does nothing. See server/device/webinspector.py.
-    "pymobiledevice3>=11.0",
+    # Web Inspector transport for simulator webview inspection. Floored at 8.0
+    # deliberately: send_plist/recv_plist are synchronous in 7.x and coroutines
+    # from 8.0 onward, so a 7.x install fails silently -- the call returns an
+    # un-awaited coroutine and does nothing. Verified at 7.7.1, 8.0.0, 9.15.1 and
+    # 11.3.0. Note this is a library import; server/device/pmd3.py separately
+    # shells out to a pipx-installed CLI, which is a different install.
+    "pymobiledevice3>=8.0",
     "pydantic>=2.0",
     "sse-starlette>=2.0",
     "mitmproxy>=10.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,10 +10,12 @@ dependencies = [
     "uvicorn[standard]>=0.34.0",
     # Web Inspector transport for simulator webview inspection. Floored at 8.0
     # deliberately: send_plist/recv_plist are synchronous in 7.x and coroutines
-    # from 8.0 onward, so a 7.x install fails silently -- the call returns an
-    # un-awaited coroutine and does nothing. Verified at 7.7.1, 8.0.0, 9.15.1 and
-    # 11.3.0. Note this is a library import; server/device/pmd3.py separately
-    # shells out to a pipx-installed CLI, which is a different install.
+    # from 8.0 onward. webinspector.py awaits them, so on 7.x the await lands on
+    # a plain return value and raises TypeError ("object NoneType can't be used
+    # in 'await' expression"). Loud, but only at call time, and the message does
+    # not mention the version -- hence the floor. Verified at 7.7.1, 8.0.0,
+    # 9.15.1 and 11.3.0. Note this is a library import; server/device/pmd3.py
+    # separately shells out to a pipx-installed CLI, a different install.
     "pymobiledevice3>=8.0",
     "pydantic>=2.0",
     "sse-starlette>=2.0",
--- a/server/device/webinspector.py
+++ b/server/device/webinspector.py
@@ -218,8 +218,17 @@ class SimulatorWebInspector:
             "WIRConnectionIdentifierKey": self._connection_id,
             "WIRApplicationIdentifierKey": application_id,
         })
-        for _ in range(8):
-            message = await self._recv()
+        # Bounded by time, not message count. The daemon interleaves unrelated
+        # reports, and a busy one can emit more than a handful before the
+        # listing arrives - a fixed iteration cap silently returns "no pages"
+        # for an app that has them. Same reasoning as _drain_handshake.
+        deadline = asyncio.get_running_loop().time() + self._timeout
+        while asyncio.get_running_loop().time() < deadline:
+            remaining = deadline - asyncio.get_running_loop().time()
+            try:
+                message = await asyncio.wait_for(self._service.recv_plist(), timeout=remaining)
+            except TimeoutError:
+                break
             if message is None:
                 break
             if message.get("__selector") == "_rpc_applicationSentListing:":
--- a/server/device/webinspector.py
+++ b/server/device/webinspector.py
@@ -1,0 +1,232 @@
+"""WebInspector — DOM access inside WKWebViews on iOS simulators.
+
+The accessibility tree does not descend into `WKWebView` on simulators, so web
+content is invisible to sim-bridge and idb. WebKit exposes a separate channel for
+this: the Web Inspector protocol, the same one Safari's Develop menu uses.
+
+On a simulator that channel is a plain unix socket — no usbmux, no lockdown, no
+pairing, no developer disk image:
+
+    /private/tmp/com.apple.launchd.*/com.apple.webinspectord_sim.socket
+
+Traffic is length-prefixed plists carrying `__selector` / `__argument` RPC pairs,
+which `pymobiledevice3.ServiceConnection` already speaks. We supply the transport
+and drive it directly rather than going through `WebinspectorService`, which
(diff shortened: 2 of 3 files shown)


jerimiah797:
Reviewed. The transport design is right and the three bugs you found by running it are the kind that only surface that way. One correction that should change the floor before this merges.

## The `>=11.0` floor is three majors too high

You asked me to confirm the coroutine boundary against my 9.15.1. I read it the other way, and measured it rather than inferring:

```
send_plist       coroutine=True
recv_plist       coroutine=True
aio_send_plist   ABSENT
aio_recv_plist   ABSENT
```

That is `inspect.iscoroutinefunction` at runtime on **9.15.1**, not a source reading. So "synchronous through 9.x" does not hold — and the explanation offered for why my PoC worked is inverted: it worked *because* I awaited coroutines, not because sync methods matched sync usage.

I then bracketed the actual transition by downloading each version from PyPI and inspecting `service_connection.py`:

| version | `send_plist` | `aio_send_plist` |
|---|---|---|
| 7.7.1 | **sync** | present |
| **8.0.0** | **async** | absent |
| 9.0.0 | async | absent |
| 9.15.1 | async | absent |
| 11.3.0 | async | absent |

The change landed at **8.0.0** — a clean major boundary, which is what you would expect for it. `aio_send_plist` was removed in the same step, so the pair moves together.

**So the floor can be `>=8.0`**, in `pyproject.toml:13` and the error string at line 155. As written the PR excludes 8.x and 9.x, which are known-good: my original PoC ran on 9.15.1 against this exact protocol and completed the handshake.

This does not change any of your conclusions — entry 14's decision stands, and your instinct to distrust "the API is identical" was correct, just resolved at the wrong version. It only widens what installs work.

Your machine at 7.7.1 does need to move regardless, since 7.x is genuinely the sync shape.

## On your three review points

**2 — declaring the dependency.** Agreed it cuts both ways, and there is a wrinkle worth deciding explicitly (logged as Q12, entry 15): on this machine `pymobiledevice3` is absent from `pyproject.toml`, not importable from `.venv`, installed as a pipx CLI at 9.15.1, and `server/device/pmd3.py` has seven subprocess call sites and zero imports. So this PR creates a *second* install — a floored library in `.venv` alongside an unfloored pipx CLI that the existing call sites keep shelling out to. Defensible, since they serve different purposes and the `--global` pipx install exists for a real reason (per 0.13.1, a per-user pipx path under an external home volume is unreachable at boot for the tunneld LaunchDaemon). But it is a choice, and #67 should record which was made rather than discover it later.

**3 — the scope boundary.** Agree, and I would keep it. Settling the protocol layer before guessing at a tool surface is the right order.

Nothing blocking beyond the floor.

jerimiah797:
### Floor lowered to `>=8.0` — you were right, and my reasoning was wrong twice

Confirmed independently before changing it. 8.0.0:

```
send_plist       coroutine=True
recv_plist       coroutine=True
aio_send_plist   ABSENT
aio_recv_plist   ABSENT
```

So the transition is at 8.0.0 and the pair moves together. `>=8.0` is now in `pyproject.toml` and the error string.

**How I got it wrong is worth recording.** I tested 7.7.1 and 11.3.0 and then *inferred* 9.x was synchronous from how your PoC was written — then stated it in entry 14 as though measured. Two versions tested, five claimed.

**And I had the mechanism backwards.** I said your PoC worked because sync methods matched sync usage. It worked because you **awaited coroutines**. That inversion is what made the wrong floor look self-consistent.

### On your Q12 — the second install

Good catch, and it was invisible to me. This PR does create two installs of the same package: a floored library in `.venv`, alongside the unfloored pipx CLI at 9.15.1 that `pmd3.py`'s seven call sites keep shelling out to.

I've made the choice explicit rather than incidental, in both places someone would look:

- `pyproject.toml` — the dependency comment now notes this is a library import and that `pmd3.py` separately shells out to a different install
- `server/device/webinspector.py` — the module docstring says the same, and that the two are floored independently

Agreed the pipx `--global` install exists for a real reason and should stay. Worth #67 recording the decision, since "why are there two" is exactly the question that costs someone an hour later.

### Scope

Agreed, keeping it. Transport only.

Nothing else outstanding from my side — this is ready for another look whenever you have one.

jerimiah797:
@coderabbitai full review

jerimiah797:
Both CodeRabbit findings look correct to me. Verified the first, since the wrong text is mine.

## The `pyproject.toml:13` comment is backwards, and that's my error

The comment says a 7.x install *"fails silently — the call returns an un-awaited coroutine and does nothing."* Measured:

```python
class Fake:
    def send_plist(self, d):   # the 7.x shape: synchronous, returns None
        return None

await Fake().send_plist({})
# TypeError: object NoneType can't be used in 'await' expression
```

**Loud, immediate, not silent.** The "un-awaited coroutine" failure is the *reverse* case — sync code calling an async method — and cannot happen when the older version is the synchronous one.

This traces back to my correction. The description came from the original entry, where it was already inconsistent with its own claim that below-the-floor was synchronous. When I moved the floor from `>=11.0` to `>=8.0` I fixed the version and left the mechanism prose alone, so it is now wrong in a way that would actively mislead whoever hits it.

Which is the point your own entry made about the first floor: a bad explanation that predicts the observed result is worse than none. I checked the boundary empirically and took the mechanism on trust.

Suggested replacement:

```
# Floored at 8.0 deliberately: send_plist/recv_plist are synchronous in 7.x and
# coroutines from 8.0 onward. On 7.x, awaiting them raises
# "TypeError: object NoneType can't be used in 'await' expression" at the first
# call — loud rather than silent, but a confusing error to land on if the floor
# is not enforced. Verified at 7.7.1, 8.0.0, 9.15.1 and 11.3.0.
```

## The `pages()` iteration cap is real

Line 221 stopping after eight messages matches what I saw independently. In my own PoC the daemon volunteered three unrelated messages before anything useful:

```
_rpc_reportCurrentState:
_rpc_reportConnectedApplicationList:
_rpc_reportConnectedDriverList:
```

Combined with your own note about applications arriving in waves, eight is a plausible number to exceed on a busy simulator — and the failure returns `[]`, which reads as "no inspectable pages" rather than "gave up early". A deadline rather than a message count seems right.

Leaving both fixes to you rather than pushing to your branch. Happy to take either if you would rather.

jerimiah797:
Both findings valid, fixed in `f175706`. The first one caught a real error I had propagated widely.

## The failure direction was backwards

You are right. This module **awaits** `send_plist`/`recv_plist`, so on 7.x — where they are synchronous — the await lands on a plain return value:

```
TypeError: object NoneType can't be used in 'await' expression
TypeError: object dict can't be used in 'await' expression
```

Loud, not silent. The silent no-op is the **opposite** mistake — synchronous code calling the 8.0+ coroutines, which yields an un-awaited coroutine and a `RuntimeWarning`. Verified both directions rather than reasoning about them.

I had that backwards in the `pyproject.toml` comment and repeated it in the commit message, the PR body, and two entries of the shared working log. The comment is corrected.

**The floor itself is unchanged and still right**, but for a better reason than I gave: the `TypeError` only arrives at call time, and its message says nothing about versions — so someone on 7.x would see a confusing await error rather than a dependency problem. A floor is more useful than that diagnostic.

## `pages()` iteration cap

Also correct, and slightly embarrassing: I had already fixed exactly this in `_drain_handshake` — the daemon interleaves unrelated reports, so a fixed cap can return "no pages" for an app that has them — and then left the same bug one function over. Now bounded by the same deadline.

Re-verified against a booted simulator after both changes: three applications enumerated, page listing unchanged. Full suite passes (1444).

I did not take the autofix option — wanted to confirm the failure semantics myself before changing the rationale, since the whole point of the comment is to explain a version constraint accurately.

Issue #40, opened 2026-06-05
Git-install `quern update` checks branch tracking, not the latest release


jerimiah797:
## Summary

For git-cloned installs, `quern update` reports "Already up to date" based on commits behind the **current branch's remote tracking ref**, not the latest tagged release. If you're on a non-`main` branch (intentionally or accidentally), the updater will happily ignore new releases and tell you everything is current. The user-visible failure mode is a silent skipped update.

## Current behavior

- **`server/lifecycle/updater.py:147` (`_update_via_git`)** is selected when `.git` exists at the project root (`_is_git_install`).
- It calls `_check_via_quern_dev(sha)` first; if that returns False, short-circuits to "up to date." Otherwise…
- …falls through to `_check_via_git(project_root)`, which does `git fetch` then counts commits behind `origin/<current-branch>` (i.e. the upstream of HEAD).
- If 0 behind → "Already up to date." If >0 → `git pull --ff-only`.

So once you're on any non-release branch — a feature branch from active development, a stale branch left from a previous session — the updater is comparing you against *your own branch's tip*, not against `main` or the latest tag. A user on `feat/foo` whose branch is in sync with origin will be told "up to date" even if main shipped three releases in the interim.

Tarball-installed users (`_update_via_tarball`, line 215) don't have this — that path hits `api.github.com/repos/quern-dev/quern/releases/latest` and does an actual release comparison. So the divergence is git-installs-only.

## Repro

1. Git-clone Quern, `./quern setup`.
2. `git checkout -b some-feature` (or `git checkout v0.13.2`, or any non-`main` ref).
3. Run `./quern update`.
4. Output: `Already up to date.` — regardless of how many releases have happened on main since the branch was cut.

Surfaced naturally today: while testing `update_quern` from inside Claude Code on the `feat/auto-update-awareness` branch (PR #39), the trigger reported "up to date" via the no-op detached-child path — which was the correct behavior *for the branch tracking semantics*, but would have been the wrong behavior if `main` had shipped a release we'd missed.

## Why this is a real snag

- The mental model for "update Quern" is "give me the latest release" — and that's what the tarball path delivers. Git users get a subtly different contract without realizing it.
- The auto-update awareness work in #39 makes this more visible: the new `update_quern` MCP tool now ends up routed through the same `_update_via_git` codepath, so an agent can be told "update Quern" and get a confidently-wrong "you're up to date."
[...]

PR #72, opened 2026-09-01
fix: reconcile Python dependencies automatically on server start

jerimiah797:
**Venv Python dependencies** now reconcile themselves. No user intervention, and the only lasting
failure is genuinely unreachable packages.

> **Scope:** this covers dependencies declared in `pyproject.toml` and installed into `.venv`.
> It does **not** cover the pipx-installed `pymobiledevice3` CLI that `server/device/pmd3.py`
> shells out to — that is a separate install, reconciled by nothing today, and tracked in #67.

## The problem

A newly declared dependency could silently fail to reach an installed server, in four different ways:

| Path | Before |
|---|---|
| `quern update`, `pip install` fails | **Warning only** — update reports success and restarts onto a stale venv |
| `quern update` returns `rc == 2` | Skips the reinstall entirely (includes the #40 feature-branch case) |
| Manual `git pull` / branch switch | Nothing runs at all |
| Server start | Nothing checked deps |

Nothing anywhere verified that the venv matched `pyproject.toml`. `check_venv()` only checks that a venv exists and is activated, not what is in it. The failure is the same shape as #67 and #68: succeeds loudly, fails quietly, and the symptom shows up somewhere unrelated much later.

The third row is the one that bites hardest in practice, since both machines on this project work on branches and pull by hand.

## The fix

Not a new pattern — the one Quern already uses for the MCP build, applied consistently. `_ensure_mcp_built` runs on every start, uses an mtime stamp against `node_modules/.install-stamp` to decide cheaply whether work is needed, and installs automatically.

`_ensure_python_deps()` mirrors it:

```
stamp   = .venv/.deps-stamp
trigger = pyproject.toml mtime > stamp mtime
action  = venv_pip install -e .
success → touch the stamp
failure → leave it alone
```

**The stamp is written only on success, and that is the whole design.** A failed install is never remembered as done, so the next start retries by itself. Fix the network, start Quern, it heals. No retry logic, no user action, no state to clear.

It covers every row in the table above, including the two `quern update` cannot reach. The in-sync case costs two `stat()` calls and touches no network.

The updater's separate `pip install` is deleted in favour of the same helper, so there is one implementation and one failure semantic rather than two.

## `quern doctor` gains a dependency section, and `--fix`

`doctor` is documented as *"read-only diagnostics"*, and that property is worth keeping — it is the tool you reach for when something is already broken, and it should not change state while you are looking at it. So the report is read-only and the repair is opt-in.

**`--fix` runs exactly what server startup runs, and deliberately not `quern setup`.** Setup does heal deps, but it has 20 `_prompt_yn` call sites, 21 `sudo` references, and at `setup.py:1605` it can `rmtree` and recreate the venv behind a prompt. Repairing a stale dependency should not be able to end in any of those — and a prompt would reintroduce exactly the user intervention this change exists to remove.

## Verified end to end, not just unit tested

- **read-only holds** — `doctor` on an out-of-sync venv reports it and writes no stamp
- **`--fix` repairs**, and a second run is a clean no-op
- **network failure**, by pointing pip at a dead index: install fails, clear message naming the network, **no stamp written**
- **automatic recovery** — the next run heals with no intervention
- the stamp lives inside `.venv`, so it is already gitignored

10 new tests, mutation-verified on both load-bearing properties:

- stamping on failure too → `test_failure_does_not_write_the_stamp` fails
- ignoring the stamp → `test_in_sync_does_no_work` fails

Full suite 1451 passed, ruff clean.

## Note

The startup call is in `_cmd_start` before `_ensure_mcp_built`, so deps are reconciled before anything that might import them. This matters for lazily-imported dependencies — PR #70 imports `pymobiledevice3` inside a function, so without this the server would start fine and only fail at first webview use.

## Summary by CodeRabbit

* **New Features**
  * Automatically synchronizes Python dependencies when the server starts.
  * Added dependency status reporting through `quern doctor`.
  * Added `quern doctor --fix` to repair missing or outdated dependencies.
  * Dependency installation failures are reported without blocking server startup.
* **Bug Fixes**
  * Dependency synchronization retries automatically after a previous failure.
  * Updates now report incomplete status when dependency installation fails.
* **Tests**
  * Added coverage for dependency detection, installation, recovery, failures, and environments without virtual environments.
* **Documentation**
  * Cleaned up the unreleased changelog section.


jerimiah797:
Reviewed from the work machine. This solves a problem I created, so first: it works, and the failure it prevents is one I actually hit.

## It fixes the concrete case

PR #70 adds `pymobiledevice3>=8.0` — the first *declared library* dependency Quern has had. During that work the pre-commit suite failed with:

```
AssertionError: assert '1 candidate' in 'pymobiledevice3 is required for webview inspection'
```

because `.venv` had never installed the newly declared dependency. I only noticed because a test asserted on a specific error string; a less specific test would have gone green against a venv missing the package entirely. That is exactly the silent path this closes.

The mtime stamp is the right shape — two `stat()` calls, no resolver, and writing it only after success so a failed install retries rather than being remembered as done. Consistent with `_ensure_mcp_built`, which is worth more than being clever.

## The gap: `pymobiledevice3` now exists twice, reconciled by neither consistently

This is your Q12 from the log, and it lands between two issues rather than inside either.

After #70 there are two installs of the same package:

| Install | Floor | Used by | Covered by |
|---|---|---|---|
| `.venv` library | `>=8.0` (declared) | `webinspector.py` (import) | **#72 — this PR** |
| pipx CLI | none | `pmd3.py`, 7 subprocess sites | #67 (report only) — **nothing reconciles it** |

So after this merges, `quern doctor` can report venv dependencies in sync while the pipx CLI sits at some unrelated version, and the two never meet. On this machine the CLI is **7.7.1** — *below* the floor #70 declares. Nothing is broken by that today, because the two paths are genuinely independent, but "the same package at two versions, one floored and one not" is the kind of thing that reads as a bug at 2am.

Not asking for it in this PR. But #72's summary line says "dependencies now reconcile themselves", and a reader would reasonably take that to include the one Quern installs via pipx. Worth a sentence in the PR body scoping it to venv dependencies, and #67 recording the pipx side as deliberately separate.

## Minor, and I think deliberate

An mtime stamp means a `git checkout` that rewrites `pyproject.toml` triggers a reinstall even when dependency content is unchanged — switching branches costs a pip run. That errs toward spurious work rather than missed installs, which is the correct direction, and matches `_ensure_mcp_built`. Flagging only so the choice is visible; I would not change it.

Nothing blocking.

jerimiah797:
Both points taken, and the first is a fair hit on my wording rather than the code.

## Scoping the claim

You're right that *"dependencies now reconcile themselves"* over-claims. It covers **venv Python dependencies only**. The pipx-installed CLI that `pmd3.py`'s seven subprocess sites use is untouched by this PR and reconciled by nothing — so after #70 there really are two installs of the same package, one floored and one not, and on your machine the CLI at 7.7.1 sits *below* the floor #70 declares.

Nothing is broken by that today because the paths are genuinely independent, but I agree it reads as a bug at 2am. I've corrected the PR summary to say venv dependencies explicitly, and Q12 in the log already records the choice; the pipx side belongs in #67 as deliberately separate rather than being quietly implied here.

That leaves an honest picture:

| Install | Floored | Reconciled by |
|---|---|---|
| `.venv` library | yes (`>=8.0`, via #70) | **this PR** |
| pipx CLI | no | nothing — #67 reports only |

## The mtime observation

Deliberate, and I'd keep it for the reason you give — it errs toward spurious work rather than missed installs, and matches `_ensure_mcp_built`. Worth noting the cost is bounded in the common case: a `git checkout` that rewrites `pyproject.toml` without changing content triggers one `pip install -e .` against an already-satisfied environment, which pip resolves quickly and locally. The alternative — hashing the dependency block instead of stat-ing the file — would avoid it, but adds parsing to a path whose whole appeal is that it's two `stat()` calls. Not worth it unless branch-switching turns out to be painful in practice.

## On the failure you hit

That's the most useful part of the review, because it's the concrete case rather than the hypothetical one: a test asserting on a specific error string caught it, and a less specific test would have gone green against a venv missing the package entirely. Worth recording that the silent version of this failure is the *normal* one — you got lucky in how it surfaced.

jerimiah797:
@coderabbitai full review

jerimiah797:
All three verified and fixed. The updater one is a good catch — it undermined the point of the PR.

## `updater.py:376` — the PR reproduced the bug it removed

`_rebuild_and_restart` discarded the boolean from `_ensure_python_deps()`, so `run_update()` still returned `0`: a successful-looking update onto a venv that had not been reconciled.

The comment I wrote **directly above that line** describes the old pip call as *"a separate pip call whose failure was only a warning, so an update could report success onto a stale venv."* Delegating to the shared helper removed the duplication and rebuilt the bug two lines below the description of it.

`_rebuild_and_restart` now returns the flag, `run_update` reports failure with a message, and two tests cover it — one asserting the failure propagates, one asserting the success path still returns `0`. Mutation-checked: discarding the return value again fails the first.

## `main.py:1051` — unreachable on a controller-less install

Correct, and the reachability argument understates it. `_cmd_doctor` exits at the `if not tools` branch, so the dependency section never ran when `/tools` reported nothing — and an installation without a working device controller is **exactly** the situation where a missing or half-installed dependency is plausible. The check was absent from the case that most needed it. Now reported on that path before the exit.

## `CHANGELOG.md:17` — mine, correct

I added a second `## [Unreleased]` when one already existed at line 8. Collapsed.

Full suite 1453 passed, ruff clean.

jerimiah797:
Re-reviewed after `4b93797`. All three CodeRabbit findings verified fixed against current head, and I went looking for a fourth in the path they did not cover. There isn't one — details below, since a negative result is worth as much here.

## The three, confirmed

**Failure propagation** — `_rebuild_and_restart` now returns `bool`, `deps_ok` is captured at `updater.py:380` and returned at `408`, and `run_update` checks it at `431` and returns `1` with *"Update incomplete: dependencies could not be installed."* That closes the loop CodeRabbit identified: previously the source updated, the venv did not, and the command announced success.

**`doctor --fix` with no controller** — fixed by calling `_report_python_deps` in **both** branches rather than moving the call. That is better than what was suggested: the comment at `main.py:1045-1047` is right that an installation with no device controller is exactly where a missing dependency is plausible, so that branch is the one that most needs the check. Moving it would have worked; duplicating it says why.

**CHANGELOG** — single `## [Unreleased]` heading now.

## I checked the startup path, and it is fine

CodeRabbit reviewed the *update* path, but this PR is named after server start, so I looked at `main.py:720`:

```python
_ensure_python_deps(quiet=True)   # result discarded
```

Discarding a result CodeRabbit had just flagged elsewhere looked like the same bug missed once. It is not:

- `_ensure_python_deps` **prints on failure regardless of `quiet`** (`__main__.py:127-134`), and the message explicitly says *"Quern will start, but features…"* — so continuing degraded is a deliberate choice, not an oversight.
- The call sits at the **top of `_cmd_start`, before the fork**, so that warning reaches the user's terminal rather than being swallowed by daemonisation.

Both of those had to be true for it to be correct, and both are. A server that refuses to start because a network blip broke `pip` would be worse than one that warns.

## One small thing, take it or leave it

`main.py:720` is now the only one of three call sites that discards the boolean, and it does so a few hundred lines from two that check it — including one CodeRabbit flagged for exactly that shape. A bare `_ = _ensure_python_deps(quiet=True)`, or a half-line comment saying failure is intentionally non-fatal here, would mark it as decided rather than overlooked. Purely a readability point; the behaviour is right.

Nothing blocking. This closes the gap that PR #70 walks straight into — a newly declared dependency that never reaches an installed server — which is the case I hit while writing it.

jerimiah797:
All three findings in the 00:23Z pass are already fixed on this branch's head — the review ran against an earlier commit.

- **Propagate dependency-install failure from the updater** — fixed. `_rebuild_and_restart` returns `bool` (`updater.py:369`), `_ensure_python_deps` result captured at `:380`, and `run_update` prints `Update incomplete: dependencies could not be installed.` and returns non-zero at `:434`. This was a real catch on the earlier commit and an embarrassing one: the PR reproduced the exact bug it exists to remove, two lines below my own comment describing that bug.
- **Report dependencies before the device-controller early return** — fixed, though not the way the suggestion proposed. Rather than moving the call above the branch, it's called on *both* paths — `main.py:1048` inside the `if not tools:` block and `:1055` on the normal path — because the early return also prints a different diagnostic line, and hoisting the call would have reordered the output so dependencies got reported before the reason for the early exit.
- **Keep one `Unreleased` heading** — this branch has exactly one. The duplicate was on #69's branch and is fixed there.

Nothing outstanding here as far as I can tell.

Issue #78, opened 2026-09-02
open_url reports success on Android when no app handles the URL


jerimiah797:
## What happens

`open_url` on Android returns `{"status": "ok"}` for a URL that nothing on the device can handle. The response is byte-for-byte indistinguishable from a successful dispatch.

Measured on a Pixel_7 AVD, Android 13, against the running daemon:

```
POST /api/v1/device/open-url  {"udid":"emulator-5554","url":"nonexistentapp12345://foo/bar"}
→ HTTP 200  {"status":"ok","udid":"emulator-5554","url":"nonexistentapp12345://foo/bar"}

POST /api/v1/device/open-url  {"udid":"emulator-5554","url":"https://example.com/"}
→ HTTP 200  {"status":"ok","udid":"emulator-5554","url":"https://example.com/"}
```

iOS gets this right, which is what makes the Android behaviour surprising:

```
POST /api/v1/device/open-url  {"udid":"<sim>","url":"nonexistentapp12345://foo/bar"}
→ HTTP 500  "[simctl] simctl openurl failed: ... (domain=NSOSStatusErrorDomain, code=-10814)
             Simulator device failed to open nonexistentapp12345://foo/bar."
```

## Why

`am start` **exits 0 even when it fails**, and writes the failure to stderr:

```
$ adb shell am start -a android.intent.action.VIEW -d 'nonexistentapp12345://foo/bar'; echo "exit=$?"
Starting: Intent { act=android.intent.action.VIEW dat=nonexistentapp12345://foo/... }
exit=0
# stderr: Error: Activity not started, unable to resolve Intent { ... }
```

`AdbController.open_url` calls `_run_adb_for_device`, which only raises on a non-zero exit code (`adb.py:752`). Nothing inspects the output — I grepped for `unable to resolve`, `Activity not started` and `No Activity` across `server/` and there are no matches.

## Why it matters more than a wrong status code

This is the failure mode we keep landing on: **a well-formed success for something that didn't happen.** An agent testing a deep link gets `status: ok`, concludes the link works, and moves on to assert on a screen that never changed — so the failure surfaces later, somewhere unrelated, as a confusing UI assertion.

It also silently undercuts the main documented use of `open_url`. Testing whether a deep link resolves is the point of the tool, and on Android it currently cannot answer that question.

## Suggested fix

[...]

PR #80, opened 2026-09-02
docs: propose fixture apps for verification and regression detection

jerimiah797:
Proposal only — no code. Prompted by #78.

## Why now

The suite mocks every subprocess, deliberately: `simctl`, `adb` and `idb` don't belong in unit tests. But that means **nothing in CI has ever executed a Quern tool against a real device.** 1467 tests prove the Python is self-consistent; they can't prove `am start` behaves the way the code assumes.

It doesn't — that's #78, present since Android support shipped, found by booting an emulator for four minutes.

Same day, one 135-line guide turned out to have four factual errors. Two were catchable by reading source. **Two were only findable by execution.**

## What I found while inventorying

Both fixture apps already exist, and the second one was a surprise:

- **QuernProbe** (`tools/probe-app/`) is in better shape than "started" — four tabs, stable identifiers everywhere, and a `selftest.py` that drives the app through the REST API rather than poking at internals. Its limit is that it carries **five assertions**, all descended from the single HID bug it was built for.

- **LogTester** is **not in version control.** It emits every iOS logging path — `print`, `NSLog`, `os_log` at four levels, `os_log` with a custom subsystem, Swift `Logger` at four levels — plus a Force Crash button, on a 2-second tick, built for both simulator and device.

  And it's load-bearing: `server/sources/device_log.py` documents its parsing regex against a sample line reading `LogTester{Foundation}[2915] <NOTICE>: ...`, dated the same day as the app's last edit. **A production parser is anchored to output from an app that exists only as a loose directory on one machine.**

## What it won't catch

Stated in the doc so this isn't oversold. Of four significant bugs this cycle, fixture apps would have caught **two** — #78 and the webview blindness. #68 and #69 are concurrency and timing; they need a load harness, and folding them in would make this proposal promise something it can't deliver.

## Proposed order

1. Bring LogTester into the repo — the only item that's a risk fix rather than an improvement
2. An Android probe — every fixture-catchable bug this cycle was on Android, and there's no Android fixture at all
3. A deep-link surface on both, including a deliberately unregistered scheme (that case *is* #78)
4. A web view tab on both, so the webview work has a fixture that isn't a third-party app
5. Run the self-tests on a self-hosted macOS runner

Three decisions are called out at the end of the doc, with recommendations. The main one: **regression detection over feature demonstration** — hostile surfaces chosen because they break tools, rather than something that looks like a real app. Demo material can come from Metatext.

## Summary by CodeRabbit

* **Documentation**
  * Added a proposal documenting real-device verification gaps, existing fixture-app coverage, and known limitations.
  * Outlined prioritized plans for expanded Android coverage, including deep-link and web-view scenarios, initially targeting emulators.
  * Proposed automated self-tests on macOS and consolidated platform probes.
  * Recorded recommendations for regression-focused, single-app-per-platform fixtures and explicit crash-test lifecycle handling.

### DIFF
--- a/docs/proposals/verification-apps.md
+++ b/docs/proposals/verification-apps.md
@@ -85,15 +85,17 @@ different tooling and should not be folded into this.
 
 Ordered by where the evidence says the holes are, not by what is easiest.
 
-1. **Bring LogTester into the repo** as `tools/log-probe/`. It is unversioned,
-   unreferenced by name, and already underpins a production regex. This is the
-   only item that is strictly a risk fix rather than an improvement, and it is
-   cheap.
+1. **Bring LogTester into the repo**, as a Logs tab and a Diagnostics tab inside
+   QuernProbe rather than as a second app. It is currently unversioned and
+   already underpins a production regex, so getting it into git is the one item
+   here that is a risk fix rather than an improvement. Emitters idle until
+   started; crash triggers behind an explicit control.
 
-2. **An Android probe app.** Every bug in this cycle that a fixture would have
-   caught was on Android, and there is no Android fixture at all. Same contract
-   as QuernProbe — stable identifiers, deterministic, offline — built with
-   Gradle rather than bare `swiftc`.
+2. **An Android probe app**, built in the combined shape from the start — UI
+   surfaces, logging surfaces and diagnostics in one artifact. Every bug in this
+   cycle that a fixture would have caught was on Android, and there is no Android
+   fixture at all. Same contract as QuernProbe — stable identifiers,
+   deterministic, offline — built with Gradle rather than bare `swiftc`.
 
 3. **A deep-link surface on both probes**, including a deliberately unregistered
    scheme. That single case is #78, and it costs almost nothing to add.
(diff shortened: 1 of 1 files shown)


jerimiah797:
Both findings were right, and both are now implemented rather than just planned — in #81.

**The log-parser assertion was the better catch.** The proposal moved LogTester into the repo and stopped there: the fixture existed, nothing asserted on the parser it exists *for*. Worse than the review knew — **nothing in the suite referenced `device_log.py` at all.** The parser behind physical-device logging shipped untested, with its only fixture an app outside version control.

`tests/test_device_log_parser.py`: 14 tests over a checked-in capture, so they run with no device attached. The probe's Logs tab emits the same shapes, which keeps a live capture and the fixture comparable and lets the probe regenerate it when the format moves.

The case worth naming is the quiet one: an unmatched line **isn't dropped**, it becomes an `INFO` entry whose message is the entire raw line with no process or level. Logs keep flowing, so a regex regression degrades silently instead of failing. One test pins exactly which fixture line is expected to miss, so a change in what falls through has to be deliberate.

Mutation-verified against three changes — subsystem made non-optional (10 failures), naive timestamps (1), a different unknown-level fallback (1).

One process note on that: my first mutation for the timestamp case reported "not caught", and the test looked weak. The mutation string had the wrong indentation and never applied. Verifying the mutation landed before trusting the result is the difference between "this test is weak" and "my check was".

**The #78 negative test already exists** in #81 and passes:

```
PASS  unresolvable scheme does not reach the app
PASS  KNOWN BUG #78: unresolvable URL still reports ok
```

The second asserts the current wrong behaviour deliberately, so it fails when #78 is fixed and gets inverted rather than deleted. The review's concern — that an unregistered scheme has no fixture screen to observe — is handled by observing the *counter's absence of movement* on the Links tab rather than looking for a screen that never appears.

Full suite: 1481 passing, up 14.

PR #81, opened 2026-09-02
feat(tools): probe apps for both platforms, with self-tests

jerimiah797:
Fixture apps for both platforms plus the runners to drive them. **iOS 11/11, Android 11/11**, verified against a booted simulator and a Pixel_7 on API 33.

```
python3 tools/run-probes.py
  PASS  iOS — QuernProbe
  PASS  Android — QuernProbe
```

The runner skips a platform when nothing is booted rather than failing it, so "no emulator running" doesn't read like a regression.

## What's here

- **`tools/probe-app-android/`** — new. Kotlin, minSdk 26, seven tabs, combined shape per the #80 decision.
- **`tools/probe-app/`** — gains Links, Logs, Web and Diag, matching Android.
- **`tools/run-probes.py`** — drives both.

**LogTester is absorbed** into the iOS Logs tab. That app was never in version control, while `server/sources/device_log.py` documents its parsing regex against LogTester's output — a production parser anchored to a fixture in a loose directory on one machine.

## Four things the fixtures found

Two are bugs in my own tests, which is the point: they were invisible until something executed.

**Android's a11y tree descends into WebView DOM.** `web_heading` and `web_paragraph` are both readable. That's the opposite of sim-bridge and `WKWebView` on iOS, so the gap we've been investigating is platform-specific rather than conceptual.

**And that exposed a false pass.** Once the page loads, the WebView container's `resource-id` disappears from the tree — only DOM nodes remain. My check asserted on the container, so it passed *only while the page was still blank*. A false pass that gets weaker the faster the machine is. It now asserts on DOM content.

**iOS shows a confirmation alert for custom-scheme links** — *Open in "QuernProbe"?* — and it blocks every subsequent query until answered. A run that died mid-test left the simulator wedged behind it, which is how I found it. Anyone automating iOS deep links needs to handle this.

**A cold-launch link arrives through `launchOptions[.url]`**, not `application(_:open:options:)`, which only fires for an already-running app. Handling one and not the other loses the link silently.

## Earlier, from the Android side

- The default `launchMode` creates a *new* activity for a VIEW intent, so every deep link reset the app to tab 0. `singleTop` fixes it; without it the fixture measured activity lifecycle while appearing to measure link delivery.
- `int(text_of(...) or -1)` turned a missing element into a value, so the #78 check passed by comparing `-1 == -1` while the tab wasn't on screen.

## Test-design notes worth keeping

`goto()` on iOS handles three real behaviours: the five-item tab bar, the More list's independent nav stack, and the fact that "More" names both a `RadioButton` tab and a `Button` back control. It also passes `scroll_to_find=False` — that defaults **on**, so asking for an absent label burns the full 60s timeout trying to scroll it into view instead of failing.

## Not included

CI wiring — item 5 in #80, wants a self-hosted macOS runner.

## Summary by CodeRabbit

* **New Features**
  * Added Quern Probe Android and iOS apps with tabbed text, controls, scrolling, logging, deep-link, web, and diagnostics surfaces.
  * Added custom-scheme and HTTPS deep-link handling with URI and receipt-count displays.
  * Added interactive controls, local web content, large scrollable lists, configurable log generation, and diagnostic actions.
  * Added Android build, installation, emulator-launch, and software-rendering support.

* **Documentation**
  * Added setup, build, emulator, troubleshooting, usage, and self-test guidance.

* **Tests**
  * Added automated Android and iOS self-tests, a runner for available devices and simulators, and expanded device-log parsing coverage.


jerimiah797:
Worked through all nine. Seven fixed, two declined with reasons.

**The most useful catch was the web assertion**, and it stings: my iOS check read `web_view present OR element_count > 0`, so a failed `goto("Web")` reported success on any non-empty screen. That's the *same false-pass shape* as the Android WebView-container check — and I introduced it two commits after fixing that one. There's now a `web_heading_native` marker that exists only on that screen, and the assertion requires it.

**The App Links gap was real coverage, not a nit.** The manifest documents an https filter deliberately without `autoVerify`, and nothing tested it: `open_url` in the runner couldn't pass a package, so only the custom scheme was exercised. The entire documented debug-build path could have broken with every check still green. Now covered and passing:

```
App Links (unverified https, package-targeted)
  PASS  package-targeted https link reaches the app
  PASS  https URI recorded
```

Also fixed: `run-probes.py` raising `FileNotFoundError` on a host without `xcrun` (taking the Android run down with it — the opposite of the skip-don't-fail behaviour it exists for), the Gradle distribution checksum pin, a duplicate Location row in the iOS README, an MD040 fence, and the author-specific `/tmp/android-jham` path.

**iOS 11/11, Android 14/14.**

---

**Declining two.**

**Password echo (`TextFragment.kt`).** `text_event_log` deliberately echoes what the app was *told*, including for `field_password`. Android does not expose a `textPassword` field's contents through accessibility, so this echo is the only way to verify typing fidelity into a secure field — and that is not hypothetical: the shift-drop bug this fixture was built to isolate was exactly a case of the field containing something different from what was sent. Redacting would remove the capability the tab exists for.

The app is offline, ships no data, and its "password" only ever holds whatever a test typed a moment earlier. I'll add a README note so it reads as a decision rather than an oversight.

**API key over cleartext HTTP (`selftest.py`).** The endpoint is `http://localhost:9100` — loopback to a local debug server, which is Quern's whole architecture. There is no TLS listener to point at, so this cannot be fixed in the self-test; it would be a change to the server's design. Worth noting the key already travels this way from every MCP call.

PR #82, opened 2026-09-02
docs: correct the web view claim; record two iOS deep link behaviours

jerimiah797:
Three corrections, all found by running the probe apps against real devices rather than by reading. The first is a published claim that is wrong on one platform.

## The web view claim is iOS-specific and was stated universally

`workflow-app-knowledge.md` — live on quern.dev — said:

> **Web view screens are a known limitation.** The accessibility tree can't see into web views. Document these prominently so agents know to use coordinate-based tapping or skip them.

True on iOS, where sim-bridge stops at the `WKWebView` container. **False on Android**, measured on API 33 against a fixture web view: `web_heading` and `web_paragraph` both come back in the tree, addressable by their DOM ids.

The advice was worse than the claim. It told Android users to abandon working identifier-based interaction for coordinates — which is slower, more brittle, and unnecessary. Now scoped by platform, with the Android behaviour stated positively.

Also fixes the interview-prompt line, which coached readers to say the same thing to a developer.

## Two iOS deep link behaviours the guide never mentioned

**iOS can ask before it opens.** A custom-scheme link raises *Open in "YourApp"?* instead of dispatching straight through, and until it's answered **every UI query returns the alert rather than the app**. Automation that doesn't expect it looks like a hung simulator. Worse, a run that dies between opening the link and answering leaves the next run failing somewhere unrelated — which is exactly how I found it.

**Cold and warm launches arrive by different routes.** Not just different app state — different entry points:

| | |
|---|---|
| Warm | `application(_:open:options:)` |
| Cold | `launchOptions[.url]` on `didFinishLaunching` |
| Universal link | `application(_:continue:...)` with an `NSUserActivity` |

Handle one and the link is lost silently on the others, with no error. A test that only runs against an already-open app never notices. The guide previously said "test cold vs warm" without saying why it breaks.

## `tap_element`'s `scroll_to_find` now states its cost

The description said "set false to fail fast without scrolling" — accurate, but it doesn't convey that leaving it on means **an element that doesn't exist is indistinguishable from one that's off-screen**, so the search runs to the ~60s request timeout and reads as a hung server.

That's the misdiagnosis pattern #68's error message exists to prevent, and it cost me an hour before I recognised it. Per `.coderabbit.yaml`, a tool description is the interface an agent reads, so this is closer to a functional fix than a docs nit.

Suite: 1467 passing. The quern.dev pages for the two guides need a re-sync once this merges.

## Summary by CodeRabbit

* **Documentation**
  * Expanded iOS deep-link testing guidance for scene-based, app-delegate, and SwiftUI apps across cold launches, warm launches, and universal links.
  * Clarified app termination steps for testing different launch paths.
  * Clarified platform differences in web-view accessibility between iOS and Android.
  * Documented `tap_element` search timeout behavior and recommended settings for presence checks.

* **Chores**
  * Added safeguards to exclude machine-specific Android configuration files from version control.


jerimiah797:
Both fixed, and the lifecycle one is worth calling out because it's the same mistake this PR exists to correct.

**Scene-based delivery.** The section described only the app-delegate lifecycle and presented it as how iOS works. It doesn't — scene-based apps, the default for UIKit since iOS 13, receive cold URLs through `connectionOptions.urlContexts`, warm through `scene(_:openURLContexts:)`, and universal links through `connectionOptions.userActivities` or `scene(_:continue:)`. An app implementing only the app-delegate callbacks never sees them fire, which is exactly the silent loss the section was warning about. Both lifecycles are now covered, plus `.onOpenURL` for SwiftUI, and the note that the app-delegate method is legacy.

The evidence was real; the generalisation wasn't. QuernProbe has no scene manifest, so the app-delegate path is genuinely what I measured — and I wrote it up as universal. That is the identical shape to the web view claim this PR corrects: measure one configuration, state it as the platform's behaviour. Two in one day, the second while fixing the first.

I've scoped the probe's own source comment too, so the fixture doesn't quietly teach the same thing.

**`local.properties`** — correct, and the mechanism is worth recording: this branch was cut from `main`, which doesn't have the probe app yet, so `tools/probe-app-android/.gitignore` doesn't exist here and `git add -A` swept up a build artifact left in the working tree. Removed, and added a root-level `.gitignore` entry so a branch predating the probe can't repeat it.

One correction to the finding, minor: it says the committed path "can use a nonexistent SDK path and fail before the probe builds." In practice `build.sh` rewrites `local.properties` from `ANDROID_HOME` on every run, so a stale committed value would have been overwritten rather than breaking a build. Still shouldn't be tracked — it's a machine-specific artifact carrying a username — but the failure mode was noise in git, not a broken build.

Issue #84, opened 2026-09-02
scroll_to_element intermittently fails to reach a distant row, and takes ~85s to say so


jerimiah797:
Separate from the no-progress fix — this reproduces on unmodified `main`.

## Measured

`tap_element(identifier="row_60")` against the probe app's 200-row list, always starting from the top of the list, on `main` at `5719e5a`:

```
run 1: HTTP 200 in 17.2s
run 2: HTTP 404 in 84.1s     ← same input, same starting state
run 3: HTTP 200 in 17.1s
```

Roughly one in three fails. `row_25` has not failed in any run, so it is distance-dependent rather than random.

## Why it matters

The failure is expensive and misleading. An agent asks for an element that **is there**, waits 84 seconds, and is told it does not exist. Retrying probably works, which is the worst shape of bug: it teaches agents to retry rather than to trust the answer, and it burns most of a minute doing it.

## Where to look

`_ios_scroll_to_element` in `server/device/controller_ui.py`. The blind sweep budget is `max_swipes * 3` = 30, with `max_swipes=10`, so there is no obvious hard limit being hit at row 60. Two candidates, neither confirmed:

- **Swipe travel varies run to run.** Successive swipes on iOS chain fling momentum, so how far the list moves depends on the interval between swipes — which depends on how long each tree read takes. A slow read early on could leave the sweep short of the target within its budget. That would explain distance-dependence and non-determinism together.
- **The visibility re-confirm** after a candidate is found (`await asyncio.sleep(0.3)` then re-fetch) rejects an over-scroll bounce. If the target lands visible and then settles out of view, the loop keeps going rather than tapping.

Worth instrumenting the loop to log swipes taken and final scroll position on failure before changing anything — the two hypotheses predict different traces.

## Not a regression

`fix/scroll-to-find-no-progress` was suspected and cleared: the same 1-in-3 failure occurs on `main` without it. I initially attributed it to that branch from a single passing `main` run, which was not enough samples to conclude anything.

jerimiah797:
Still present on **0.18.2**, and reproduced from a second entry point with a
larger budget — both of which bear on the hypotheses above.

## Reproduced through the dedicated endpoint, not only `tap_element`

The measurements in the issue go through `tap_element(..., scroll_to_find=True)`.
This is `POST /api/v1/device/ui/scroll-to-element` — the call whose whole job is
the scroll — against the same fixture (`tools/probe-app`, 200 rows) and the same
target, `row_60`, always from the top of the list:

[...]

PR #85, opened 2026-09-02
fix(ui): stop the blind scroll sweep when nothing is scrolling

jerimiah797:
Fixes the reported symptom: `tap_element` swiping repeatedly on screens with no scrollable content, which makes Quern look broken.

## Root cause

The stall guard already existed — on the wrong branch. When the target is in the tree but off-screen, the loop correctly detects the container has hit its end and aborts. When the target is **absent entirely** (a typo, a wrong label, an element on another screen) it falls into the blind sweep:

```python
else:
    # Not located (recycled/lazy row): blind sweep, down then up.
    y1, y2 = (y_far, y_near) if blind_steps < blind_down else (y_near, y_far)
    blind_steps += 1
    last_cy = None
    stalls = 0        # ← reset every iteration; nothing checks for progress
```

`max_swipes * 3` = **30 swipes**, with nothing asking whether the screen ever moved.

## Fix and measurement

A single fingerprint of on-screen geometry, taken around the first blind swipe. A swipe over scrollable content always moves geometry, so identical positions mean nothing scrolled and the remaining budget would repeat that no-op.

**Static screen, missing element: 16.6s → 12.5s**, and the visible thrashing stops. Tests are mutation-verified: removing the check restores the full-budget sweep and fails.

## Two things I tried and dropped, both because measurement contradicted them

**Sharing one unfiltered tree read** between the progress check and the target lookup — an apparently free halving of per-iteration cost. It broke scrolling: a target 60 rows down went from found to not-found. A filtered read is **not** the unfiltered read minus other elements, so client-side matching over the full tree isn't equivalent. Reverted.

**Seeding the fingerprint from the caller's element list**, to make the first comparison free. `tap_element`'s read is filtered to the missing target, so that list is always `[]`. It did nothing. Removed rather than left as dead API surface — an existing test caught it by asserting the call signature.

## One correction on my own reasoning

I reported the row_60 failure as a regression from this branch, based on **one** passing run of `main`. Sampling `main` three times shows it fails 1 in 3 on its own:

```
main run 1: HTTP 200 in 17.2s
main run 2: HTTP 404 in 84.1s
main run 3: HTTP 200 in 17.1s
```

Pre-existing, not this branch, and filed separately as **#84** with the two candidate mechanisms and a note that neither is confirmed. Concluding from a single sample is the exact error I've been flagging elsewhere today.

## Scope

This does not change the `scroll_to_find` default. That was the original proposal, and the measurement argues against it: the default is useful, and the defect was the loop not noticing it was making no progress. Suite: 1485 passing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved iOS scrolling when searching for content, including screens positioned at the bottom of a scrollable view.
  - The app now checks the reverse direction before determining that a screen cannot scroll further.
  - Scrolling stops early when the screen remains unchanged, reducing unnecessary swipe attempts.
  - Visible content is detected immediately without an extra swipe.
  - Improved handling of visual changes that do not represent scrolling progress.
  - Improved reliability when confirming that a searched-for element is not present.

### DIFF
--- a/server/device/controller_ui.py
+++ b/server/device/controller_ui.py
@@ -465,8 +465,23 @@ class DeviceControllerUI:
             if el is None and not progress_checked and blind_signature is not None:
                 progress_checked = True
                 signature = await _signature()
-                if signature:
-                    if blind_signature is not None and signature == blind_signature:
+                if signature is not None:
+                    # An unchanged screen after a *downward* swipe is not enough
+                    # to call it static: a list already scrolled to the bottom
+                    # cannot move further down while content above it is still
+                    # reachable. Probe the other direction before concluding,
+                    # otherwise targets above the viewport report not_found.
+                    if signature == blind_signature:
+                        await _swipe(y_near, y_far)
+                        reverse = await _signature()
+                        if reverse is not None and reverse != blind_signature:
+                            blind_signature = reverse  # it moves; carry on
+                            el = await _fetch()
+                            if el is not None and _visible(el):
+                                return el
+                            continue
+                        signature = blind_signature  # neither direction moved
+                    if signature == blind_signature:
                         # A swipe over scrollable content always moves geometry.
                         # Identical positions mean nothing scrolled, so the
                         # remaining budget would repeat this exact no-op.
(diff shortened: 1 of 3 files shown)


jerimiah797:
@coderabbitai review

PR #86, opened 2026-09-02
chore: gate merges on a current review

jerimiah797:
CodeRabbit re-reviews on **every push**, so a "0 unresolved threads" reading taken before the latest push is stale — and it looks identical to all-clear. Merging on that means merging unreviewed code. That happened more than once today.

## Why not a post-push hook

The obvious fix is a reminder after pushing, and I don't think it's the right one:

- Git has no `post-push` hook. `pre-push` fires *before* the push, so the review it would remind you about cannot exist yet.
- A reminder only tells me something I already knew. Knowing wasn't the failure.
- **Pushing without checking is harmless.** The risky moment is the *merge*.

So the check lives at the merge, where acting on stale data actually costs something.

## What this adds

```sh
scripts/pr-review-status.py            # every open PR
scripts/pr-review-status.py 85 --wait  # block until current
scripts/merge-pr.sh 85                 # refuses unless the review is current
scripts/merge-pr.sh 85 --wait          # waits, then merges
scripts/merge-pr.sh 85 --force         # override, deliberately
```

`pr-review-status.py` compares the newest review timestamp against the newest commit and exits non-zero unless the PR is **both** current and clean.

Verified refusing on a genuinely stale PR:

```
WAIT #85 fix(ui): stop the blind scroll sweep when no — pushed since the last review

Not merging #85: its review is not current, or it has unresolved findings.
```

The rule is in `CLAUDE.md`, since a script only helps if it gets reached for — and reaching for it is the part that failed.

## Honest limit

This gates *my* merges through the script. It doesn't stop anyone running `gh pr merge` directly. Branch protection requiring an approving review would enforce it properly; this is the cheap version that fixes the observed failure.

## Summary by CodeRabbit

- **New Features**
  - Added pull request review-status checks for pending, unresolved, and indeterminate feedback.
  - Added optional polling until reviews are complete.
  - Added merge safeguards that block unresolved or uncertain reviews, with an explicit force option to bypass checks.
  - Merges now verify that the pull request’s latest revision is being merged.

- **Bug Fixes**
  - Review checks now fail safely when status information is unavailable or incomplete.

- **Chores**
  - Automatic incremental reviews are disabled after the initial review; subsequent reviews must be requested manually.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -20,6 +20,14 @@ reviews:
   auto_review:
     enabled: true
     drafts: false
+    # Review when a PR opens, then only when asked. Re-reviewing every push
+    # sounds strictly better and is not: review runs are capped per hour, and
+    # fixing findings means pushing, so the automatic behaviour spends the
+    # budget on intermediate states nobody merges and leaves the final one
+    # queued behind them. scripts/merge-pr.sh asks for a review at the point it
+    # matters -- immediately before a merge -- so the cap is spent on the commit
+    # that actually ships.
+    auto_incremental_review: false
 
   path_filters:
     # Generated or vendored — reviewing these produces noise, not findings.
--- a/scripts/merge-pr.sh
+++ b/scripts/merge-pr.sh
@@ -7,9 +7,11 @@
 # risky moment is the merge, not the push -- so the check lives here, where
 # acting on stale data actually costs something.
 #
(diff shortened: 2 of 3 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #88, opened 2026-09-02
fix(scripts): refuse while GitHub's view of the head is stale

jerimiah797:
**The gate merged an unreviewed commit.** Reporting that plainly because it's the third time this tool has failed open.

## What happened

```
22:11:39  commit pushed
22:11:40  gate ran → "reviewed, clean"
22:11:45  merged
22:09:30  ← the review it trusted, two minutes older than the commit
```

The check ran **one second** after the push. GitHub hadn't registered the new commit yet, so `gh pr view --json commits` returned the *previous* head — whose review genuinely was current. No review was ever requested; the comment count never moved.

The staleness was invisible because every timestamp involved was internally consistent. Only the head SHA disagreed, and nothing was looking at it.

## The fix

Compare the API's `headRefOid` against `origin/<branch>`. A mismatch raises, which reports `unknown` and **refuses** — `unknown` has never been a pass. `merge-pr.sh` now runs `git fetch` first so the comparison means something.

Simulated with a stale head:

```
state=unknown
detail=the API still reports head 00000000 while origin/main is ac9114d4 — it has not caught up with the push
exit would be 4 (0 would merge)
```

## The pattern worth noting

Three fail-opens in one small tool:

1. `gh` errors treated as data → empty result read as "no commits" → any prior review looked newer
2. the summary comment mistaken for a review signal → refreshes on every push
3. an API read racing a push → previous head's review accepted for the new commit

Every one produced a confident "reviewed, clean". None produced an error. For something whose entire purpose is refusing, the consistent failure direction is worth keeping in mind — it fails toward yes.

**No damage from the premature merge**: `main` is green at 1486 tests and lint-clean, and the merged commit was the review-handshake work, unreviewed but not wrong.

## Summary by CodeRabbit

- **Improvements**
  - Review-status checks now confirm that the pull request’s latest commit is available locally before reporting success.
  - Merge operations now use the verified latest pull request commit, helping prevent merges based on stale information.
  - Added safeguards when the latest commit cannot be verified.
  - Forced merges can proceed using the current pull request commit when verification is unavailable.
  - Review-status checks now report an unknown status instead of exposing an execution error when Git operations fail.
  - Temporary verification data is cleaned up automatically after operations complete.

### DIFF
--- a/scripts/merge-pr.sh
+++ b/scripts/merge-pr.sh
@@ -38,7 +38,9 @@ else
   # review object, and the summary comment refreshes on every push whether or
   # not a review ran. Asking costs a comment, which is why the plain status
   # check stays read-only and only the merge path pays it.
-  python3 scripts/pr-review-status.py "$PR" --ask $WAIT || STATE=$?
+  HEAD_FILE=$(mktemp)
+  trap 'rm -f "$HEAD_FILE"' EXIT
+  python3 scripts/pr-review-status.py "$PR" --ask --emit-head "$HEAD_FILE" $WAIT || STATE=$?
 
   case "${STATE:-0}" in
     0) ;;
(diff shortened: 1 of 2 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

Issue #66, opened 2026-09-01
Running XCUITest on a simulator kills the accessibility tree for all foreground apps until reboot


jerimiah797:
## Summary

After **any** XCUITest run against a simulator, `get_ui_tree` / `get_screen_summary` return a single bare `Application` element with a `0x0` frame for **every foregrounded app** — including apps that were never under test. Only SpringBoard still reports normally. A simulator reboot is the only known fix.

This makes a simulator unusable for Quern after a UI-test run, and the empty tree looks exactly like mass landmark drift, so it is easy to misdiagnose.

## Reproduction

Deterministic. Reproduced from clean boots, multiple times, on two runtimes.

1. Boot a simulator, launch an app, confirm the tree is healthy
2. Run **one passing** XCUITest against that simulator
3. Relaunch the app and read the tree

```
xcodebuild -workspace <ws> -scheme "<UI scheme>" -sdk iphonesimulator \
  -destination 'id=<UDID>' -only-testing:<Target>/<Class>/<oneTest> test
```

A crash or abnormal exit is **not** required — a clean `** TEST SUCCEEDED **` poisons it.

## Observed

| Target | Healthy | After one XCUITest run |
|---|---|---|
| App under test | 20 elements, `confidence: exact` | **1** — bare `Application`, `frame 0x0`, `label: null` |
| **Safari** (never under test) | normal | **1** — same signature |
| SpringBoard | 12 | **12 — unaffected** |

## Affected versions

Not runtime-specific:

| Device | Runtime | Healthy | After XCUITest |
|---|---|---|---|
| iPhone 16 Pro | iOS 18.6 (22G86) | 13 | 1 |
| iPhone 17 Pro | iOS 26.5 (23F77) | 11 | 1 |

Xcode 26.5 (17F42).

## It is not Quern

Raw `idb ui describe-all`, with Quern entirely out of the picture, shows the identical signature. sim-bridge is not the cause.

> ⚠️ When checking this by hand, **count the parsed JSON array — not `wc -l`**. idb emits the whole array on one line, so `wc -l` reports `1` even on a healthy simulator and looks like a false positive.

```bash
idb ui describe-all --udid <UDID> | python3 -c "import json,sys; print(len(json.load(sys.stdin)))"
```

[...]

PR #89, opened 2026-09-03
fix(device): recover the accessibility bridge wedged by XCUITest (#66)

jerimiah797:
Closes the last step of #66. The investigation there was already conclusive; what was missing is that the fix lived as a command to remember rather than something Quern does.

## The problem

Any XCTest-based run against a simulator — XCUITest **or WDA** — leaves that simulator's `CoreSimulatorBridge` holding a stale mach-port cache. Every foregrounded app then reports a single bare `Application` element with a `0x0` frame. Not just the app under test: **Safari breaks too**, and only SpringBoard keeps reading normally.

What makes it worth automating is the misdiagnosis. An empty tree is indistinguishable from every landmark on every screen drifting at once, so the natural response is to go and fix knowledge bases that were never wrong — and to suspect the bridge you'd have to already know about this.

## What it does

Recognise the signature, restart that simulator's bridge, retry once. The bridge is launchd-on-demand, so the retried query comes back against a fresh cache.

- **~1s**, app state preserved
- **Scoped to one simulator** — there's one bridge per booted simulator, and `lsof` maps it to the UDID. The others are undisturbed.
- **Once per read.** If the tree still looks wedged afterwards, the cause is something else — and a retry storm is actively harmful here, since sim-bridge serialises commands without cancelling abandoned ones, so repeated attempts drain into something that presents as a hang (#68).

## Detection is deliberately narrow

An app mid-launch legitimately reports a single `Application` element. The **zero frame and absent label** are what separate "nothing has rendered yet" from "the bridge cannot see anything at all", and a false positive costs a needless restart. Tested both ways.

## Verification

Nine tests, mutation-verified against two plausible regressions — killing every bridge instead of the matching one, and dropping the zero-frame check.

Kill and respawn verified live rather than only mocked:

```
bridge pids for F5AF3736: [28920]     ← correctly matched to the UDID
reset_bridge -> True
bridge after query: 91862             ← launchd respawned it
fresh read: 10 elements in 2.25s      ← healthy
```

One note on that verification: my first attempt "confirmed" recovery with a query that returned in 0.05s — which was the UI cache, not the bridge. It only became possible to be fooled that way because the cache timestamp fix earlier today made the cache actually serve reads for the first time. Worth knowing generally: **a test that reads twice in quick succession may now be measuring the cache rather than the thing under test.** The numbers above are with `use_cache=false`.

Suite: 1495.

## Not included

A manual `reset_ax_bridge` tool, suggested in the issue as a follow-on. Auto-recovery makes it unnecessary for the common case; it would only earn its place if a partially-wedged state turns up that detection misses.

## Summary by CodeRabbit

- **New Features**
  - Improved simulator accessibility recovery after XCUITest/WDA runs.
  - Automatically refreshes a stale accessibility state once for the affected simulator.
  - Preserves the app state during recovery and limits changes to the relevant simulator.

- **Bug Fixes**
  - Prevents stale accessibility trees from blocking simulator interaction.
  - Retains healthy accessibility results after recovery.
  - Avoids repeated recovery attempts when accessibility data remains unavailable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to Quern are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- **A simulator no longer stays unusable after an XCUITest or WDA run** (#66) — any XCTest-based run leaves that simulator's `CoreSimulatorBridge` holding a stale mach-port cache, after which every foregrounded app reports a single bare `Application` element with a `0x0` frame. Not just the app under test: Safari breaks too, and only SpringBoard keeps reading normally. The empty tree is indistinguishable from every landmark on every screen drifting at once, so the usual response was to go and edit knowledge bases that were never wrong. Quern now recognises the signature and restarts that simulator's bridge, which is launchd-on-demand, so the retried query comes back against a fresh cache. Around a second, app state preserved, and scoped to the one simulator — there is one bridge per booted simulator, so the others are undisturbed. Recovery is attempted once per read; if the tree still looks wedged afterwards the cause is something else, and retrying would only be a slower way to return the same answer.
+
+### Notes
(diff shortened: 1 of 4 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #90, opened 2026-09-03
feat(device): reach web content inside WKWebView

jerimiah797:
Quern could not see inside a `WKWebView` on iOS simulators. The tree walk stops at the web view — which is not merely empty in the tree, it is **absent entirely** — so a screen built on web content looked like native chrome and nothing else. That shaped real decisions: the earlier investigation concluded WDA was the only route, `workflow-app-knowledge.md` told readers to fall back to coordinate tapping, and a `tap_element` against a web page burned 55 seconds sweeping for something structurally invisible.

Two routes now work. Which applies is decided by the app, not by us.

## Route 1 — the Web Inspector (preferred)

`webinspector.py` merged unwired; this is the half that returns something. Against the probe app:

```
<h1>     id=web_heading    (24, 24 354x39)
<button> id=web_button     (24,120  85x20)  INTERACTIVE
<span>   id=web_nested     (24,160 142x20)
```

Real DOM ids, complete enumeration, an interactivity flag, and JS evaluation. **15ms.**

The protocol detail that cost the most: messages must be wrapped in `Target.sendMessageToTarget`, with replies inside `Target.dispatchMessageFromTarget`. A bare `Runtime.evaluate` is answered with *"'Runtime' domain was not found"*, which reads as "this WebKit has no JavaScript" rather than "you skipped a layer".

Verified on a **third-party page in a third-party app** — Metatext's instance picker, `joinmastodon.org/servers` — after a one-line debug-build change. It found the icon-only `Toggle menu` button, which has no text and is therefore invisible to every screenshot- or accessibility-based approach.

## Route 2 — accessibility hit-testing (fallback)

`describe_point` reaches web content even though the tree walk does not. Works on release builds and on apps that never opted in. Three targeting strategies were measured; two were discarded:

| | probes | time | found |
|---|---|---|---|
| blind lattice | 57 | 5.3s | 2 of 4, on a *simpler* page |
| nearest-hint following | — | **never terminated** in 300s | — |
| Vision-guided (`VNRecognizeTextRequest`) | **4** | **829ms** | 4 of 4 |

Vision runs in-process through pyobjc — no compile step, nothing to sign or ship. Language correction is off: it rewrites UI strings into dictionary words, and the text is matched against accessibility labels.

## The two routes compose

DOM geometry is viewport-relative and so not directly tappable. One hit-test anchors it: the heading reads `(24,24)` in the page and `(24,124)` on screen, and predicting the button's centre from the DOM then hit-testing there returns `Button "Web Button"`. The probe work is the calibration for the Inspector, not a competitor to it.

## Knowledge base

Screens gain a `web_content:` block — host, `in_process`, `inspectable`, `url`, `page_offset`. These are facts only the source can answer. `in_process` decides everything else: an app that constructs the view can opt it into inspection; one handed an `SFSafariViewController` cannot.

Also corrects `workflow-app-knowledge.md`, which still told readers to fall back to coordinates on iOS.

## Things worth knowing

- **A hit-test in whitespace returns the nearest element, not `None`.** `hit_contains` is the validity test; without it a sweep records elements that are not where it says they are. The fixture reproduces this deliberately.
- **Edge detection was the wrong primitive** for the pixel fallback: an edge filter marks a shape's *border*, so a solid-filled button arrives as two hairlines as far apart as the button is tall. Text hid it until a synthetic image exposed it. It uses difference-from-background now.
- **`probing.py` is untouched** beyond promoting `_frame_key` to a public `frame_key`, with the private name kept as an alias, so tab-bar probing on the hot `describe_all` path is byte-identical.
- OCR is lossy (`'fMastodon'`) — but it only *aims*; identity comes from the hit-test, so an error costs a mis-aimed probe rather than a wrong label.

## Not included

The route and MCP tool. This is the capability; the surface is a separate change, and worth deciding after review given the Inspector turned out to dominate the probe for the primary use case.

Suite: 1516.

## Summary by CodeRabbit

- **New Features**
  - Added support for inspecting and extracting visible content from web-based screens.
  - Improved discovery of web content that is not exposed through standard accessibility data.
  - Added macOS Vision-based text recognition with fallback image analysis.
  - Added optional web-content configuration for documenting screen behavior.

- **Documentation**
  - Expanded guidance for iOS and Android web views, inspectability, URLs, accessibility, and page offsets.
  - Added examples and recommendations for documenting web-backed screens.

### DIFF
--- a/docs/app-knowledge-guide.md
+++ b/docs/app-knowledge-guide.md
@@ -243,6 +243,52 @@ This is a separate issue from identifier reliability. The label is *correct*, bu
 - **In the Key Elements table**, write the stable portion followed by `...` and explain the dynamic part in Notes. Example: `Attributes, ...` with note "Comma-separated list of cache attributes, varies per cache."
 - **Prefer identifier over label** for dynamic-label elements — this is the one case where identifiers are more reliable than labels, if one exists.
 
+### Documenting Web Views
+
+A screen built on web content looks almost empty to `get_ui_tree`: on iOS the
+accessibility tree does not descend into a `WKWebView`, and the out-of-process
+hosts are not in the app's hierarchy at all. The failure mode is a
+misdiagnosis, not an error — a near-empty tree reads as "the screen failed to
+load" or "every landmark drifted", and the agent goes looking for a problem
+that isn't there.
+
+Record it in the screen's `web_content:` block. The facts that matter are ones
+only the source can answer, which is exactly what a knowledge base is for.
+
+```yaml
+web_content:
+  - host: "WKWebView"
+    in_process: true
+    inspectable: "debug"
+    url: "https://joinmastodon.org/servers"
+    page_offset: [0, 100]
+```
+
+**`in_process` decides everything else.** When the app constructs the view, it
+can opt it into remote inspection and the page's own DOM becomes available.
+When the system constructs it — `SFSafariViewController`,
+`ASWebAuthenticationSession` — the app has no handle on it, so that route is
+closed no matter what the app does.
+
+**`inspectable` is about the view, not the content.** Since iOS 16.4 a
+`WKWebView` is only inspectable if the app sets `isInspectable` on it. That is
+the app's decision about a view it owns, so it can be true even when the HTML
+comes from a third party the team has no control over. Use `"debug"` when it is
+behind `#if DEBUG`, which is where it belongs.
+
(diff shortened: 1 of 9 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #91, opened 2026-09-03
feat: reach web content inside WKWebView from an agent

jerimiah797:
Follows #90, which landed the Web Inspector plumbing unwired. This makes it
reachable: `get_web_content` returns a page's elements with real screen frames,
so an agent can read and tap web content that `get_ui_tree` cannot see at all.

## The problem this solves

A `WKWebView` is **absent from the accessibility tree on iOS simulators** — not
an empty container, not a leaf. A screen built around one looks like nothing but
its native chrome, which is how a `tap_element` against Metatext's instance
picker once burned 55 seconds sweeping for something structurally invisible.

## Anchoring

The Inspector reports geometry in page space; Quern speaks device points. The
offset is measured, not assumed: the difference between an element's observed
accessibility frame and its known page rect fixes the whole page.

The first design searched for any DOM text matching any probe hit. **It is not
merely slower but wrong.** On joinmastodon.org a logo whose `aria-label` is
`"Mastodon"` is a prefix of the body paragraph `"Mastodon is not a single
website..."`; pairing them anchored the page 215pt from where it sat —
confidently, with no error raised anywhere. What ships instead hypothesises an
origin from geometry the chrome gives away, then confirms one *predicted*
element. Asking "is this the element I expected here" cannot make that mistake.
`tests/test_web_content.py` records it as a regression.

## Two collector fixes, both found by measuring

- **Block box vs. rendered glyphs.** A block element's border box spans its
  whole column while accessibility reports only the text, so probing the box
  centre lands in whitespace and answers with the nearest element. A `Range`
  over the text nodes reproduces the AX frame exactly — an `<h1>` with a 354x50
  box has a text rect and an AX frame that both read 144x53. This applies to
  controls too: a nav link whose `<a>` spans a 394pt row reports a frame only as
  wide as the word "Apps".
- **Occlusion.** An element under an overlay still has a valid box. With the
  site's nav menu open, the heading and paragraph beneath it were reported as
  tappable; `elementFromPoint` now drops them.

## Verification

Live against Metatext's instance picker, projecting every element and
hit-testing its centre:

| state | elements | result |
|---|---|---|
| menu closed | 4 | **4/4** land on their own element |
| menu open | 8 | **8/8** |

A tap at a reported frame opened the nav menu — page 3 went from 4 elements to
10 — which is the end-to-end proof that the frames are real.

Cost is ~2.4s warm, of which **~1.8s is the pre-existing native tree walk**
(23ms when its cache is warm). The inspector and probes account for ~0.5s.

## Notes for review

- 26 new tests, each mutation-verified: every guard was individually broken and
  the intended test confirmed to fail. (Watch for stale `.pyc` when doing this —
  a same-length mutation restored within the same second reuses the mutant's
  bytecode, which cost me a bogus failure.)
- The three JS changes in `_COLLECT_JS` are covered by live measurement rather
  than unit tests; there is no JS engine in the suite.
- **Known limitation:** the `webinspectord` socket carries no UDID, so with more
  than one simulator booted the response warns that the content may belong to
  another device. Also, the server now holds a persistent inspector connection,
  which conflicts with a second client on the same socket.
- Not in scope: merging these into `get_ui_tree` / `tap_element` via the overlay
  and pre-tap verification from the plan. Frames are tappable today via `tap`.

Suite 1544 passing, ruff clean, MCP builds.

## Summary by CodeRabbit

* **New Features**
  * Added `get_web_content` support for reading live WKWebView content on iOS simulators.
  * Added HTTP and MCP access, with optional device and app identifiers.
  * Results include web-page elements, accessibility details, and screen coordinates.
  * Supports coordinate-based tapping using returned element frames.

* **Bug Fixes**
  * Improved geometry, text bounds, filtering, and overlay handling.
  * Added safeguards to prevent results from being returned for the wrong simulator.
  * Improved reliability when multiple simulators or web pages are active.

* **Documentation**
  * Updated tool documentation and the listed tool count.


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #92, opened 2026-09-03
feat: make web content tappable by label

jerimiah797:
Follows #91, which shipped `get_web_content` with a wart: because the elements
existed only in that tool's response, its description had to tell agents **not**
to use `tap_element` — the one instruction most likely to be ignored, and whose
failure mode is the 55-second scroll sweep this whole line of work exists to
remove.

## What changed

Web elements are kept as a per-device overlay and merged into UI reads **on
return**, so `tap_element`, `get_ui_tree`, `get_screen_summary`, landmark
matching and annotated screenshots all see them. The merge is on the way out
rather than into the cache, so the cached native tree stays purely native.

## Staleness is the whole problem

A web element's coordinates describe the page as it was when it was read, and
**nothing in the native tree will ever contradict them** — the tree cannot see
the page at all. So a stale overlay is confidently wrong in a way no later read
detects. Two layers:

- The overlay is dropped wherever the UI cache is. That is one choke point every
  mutating operation already funnels through, so a tap, swipe, scroll, launch or
  device change all clear it without needing to be enumerated.
- A tap against a surviving entry is **probed first**. If the point no longer
  holds that element, the overlay is dropped and `tap_element` returns
  `not_found` with `reason: "stale_web_content"` — one ~93ms probe and an honest
  error instead of a tap on whatever now occupies that pixel.

An unlabelled control (an icon-only button) is confirmed by position alone;
there is no text to compare, and refusing would make it untappable.

## A deliberate omission

A DOM id is **not** exposed as an accessibility identifier. It never appears in
the native tree, so letting it match `identifier=` would make a web-only lookup
indistinguishable from an ordinary one — and silently unresolvable the moment
the overlay expires.

## Verification

Live on Metatext's instance picker:

1. `get_web_content` → 4 elements, anchored.
2. `get_screen_summary` → *"Interactive elements: Done, Forward, **Toggle menu**, Back"*.
3. `tap_element(label="Toggle menu")` → `ok`, `source: web-inspector`; the site's
   nav menu opens.
4. The same call again → **404**, screen context back to 3 native buttons, the
   overlay having been dropped by the tap in step 3.

17 new tests, each mutation-verified — every guard broken individually and the
intended test confirmed to fail. Suite 1569 passing, ruff clean, MCP builds.

## Note

The stale-overlay path cannot be exercised live, because every quern-driven
change drops the overlay through the same choke point; reaching it requires the
page to change on its own (a timer, animation or late network response). It is
covered by unit tests instead.

## Summary by CodeRabbit

* **New Features**
  * Web content extracted through `get_web_content` now appears in subsequent UI reads and can be interacted with using `tap_element`.
  * Web elements are filtered, deduplicated, and retained for up to 60 seconds.
  * Taps verify that web content is still current, reporting when a page has changed or is no longer visible.
* **Bug Fixes**
  * Prevented outdated web content from affecting subsequent page reads.
* **Documentation**
  * Updated UI element attribute documentation for Android and web content metadata.


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #93, opened 2026-09-03
feat: reach web views the host app does not own

jerimiah797:
Found by driving a real OAuth login against a self-hosted GoToSocial instance.
Follows #91 and #92.

## The gap

`SFSafariViewController` and `ASWebAuthenticationSession` run **out of process**.
`isInspectable` on the host app's own web views has nothing to do with them, and
they carry their own chrome — an address bar above, a toolbar below — none of
which appears in the native accessibility tree.

So both existing candidate origins fail:

- *"below the top chrome"* — there is no chrome in the tree to measure.
- *"as tall as its viewport, anchored to the foot of the screen"* — wrong by the
  height of the bottom toolbar.

Measured on GoToSocial's settings page: a **402x685 viewport on a 402x874
screen sits at y=106**, while bottom-anchoring predicts **189**.

## The fix

When geometry suggests no origin that holds, sweep for one: probe a few columns,
match hits to DOM text, take the largest cluster of agreeing offsets.

Two constraints make sweeping safe rather than merely cheap, and both are
regression-tested:

- **Exact matching only.** The design deleted in #91 accepted containment and
  paired a logo labelled `"Mastodon"` with a paragraph beginning `"Mastodon is
  not a single website…"`, placing the page 215pt from where it sat —
  confidently, with no error anywhere.
- **Two independent probes must agree.** A page and a screen can share a string
  by coincidence; one match is not evidence.

## Two label fixes from the same page

- **Form controls.** Accessibility names an input from its associated `<label>`,
  not its value. Collecting the value made every input fail comparison — an
  `<input>` holding `https://social.arctian.org` is reported as `"Instance"`.
- **Flattened nesting.** An emoji `<span>` inside a link answers as the link, so
  a probe landing on a **strictly larger enclosing** element now confirms.
  Strictly larger matters: equal frames say nothing about nesting, and accepting
  them would let an element swapped in at the same coordinates pass for the one
  that used to be there. That mutation is covered.

## Verification

Metatext's Account Settings, where **the native tree reports 1 element**:

| route | result |
|---|---|
| accessibility tree walk | 1 element (the Application) |
| Web Inspector alone | 13 elements, but **unanchored** — no origin |
| with the sweep fallback | **13 elements, anchored**, 13 probes, ~3.2s |

Every projected frame matches a position measured independently by hit-test
(y = 145, 252, 325, 543, 599, 679, 735).

## Two boundaries worth recording

- `SFSafariViewController` **is** exposed to the Web Inspector as
  `com.apple.SafariViewService`, with no opt-in from the host app.
- `ASWebAuthenticationSession` is **not** — during a live OAuth flow the
  inspector reported *zero* connected applications. Hit-testing still reaches it,
  which is how the sign-in below was driven.

The existing "several apps are connected … pass bundle_id" message did its job
here rather than guessing between Metatext and SafariViewService.

Suite 1579 passing, ruff clean. New guards individually mutation-verified —
including one test that initially passed for the wrong reason (its element was
narrower than the sweep columns, so nothing ever matched).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved web-element validation when accessibility overlays enclose elements, even when labels differ.
  - Improved web-view positioning by checking all pages before fallback probing, requiring distinct agreeing matches, and respecting probe limits.
  - Improved form-control information by recognizing associated labels, ARIA labels, and placeholders while retaining field values.
  - Improved inspector connection cleanup to prevent stale sessions and ensure reliable shutdown.

- **Tests**
  - Added coverage for positioning, overlay verification, and inspector connection cleanup.

### DIFF
--- a/server/device/controller_ui.py
+++ b/server/device/controller_ui.py
@@ -730,6 +730,30 @@ class DeviceControllerUI:
         else:
             self._web_overlay.pop(udid, None)
 
+    @staticmethod
+    def _encloses(outer: dict | None, inner: dict | None) -> bool:
+        """Whether `outer` strictly contains `inner`.
+
+        Strictly: two frames of the same size say nothing about nesting, and
+        accepting them would let an element swapped in at the same coordinates
+        pass as the one that used to be there -- which is the staleness this
+        check exists to catch.
+        """
+        if not outer or not inner:
+            return False
+        outer_area = outer.get("width", 0) * outer.get("height", 0)
+        inner_area = inner.get("width", 0) * inner.get("height", 0)
+        if outer_area <= inner_area:
+            return False
+        return (
+            outer.get("x", 0) <= inner.get("x", 0)
+            and outer.get("y", 0) <= inner.get("y", 0)
+            and outer.get("x", 0) + outer.get("width", 0)
+            >= inner.get("x", 0) + inner.get("width", 0)
+            and outer.get("y", 0) + outer.get("height", 0)
+            >= inner.get("y", 0) + inner.get("height", 0)
+        )
+
     async def _web_element_still_there(self, udid: str, element: UIElement) -> bool:
         """One probe, before tapping something the tree cannot see.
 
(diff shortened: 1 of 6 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #94, opened 2026-09-03
feat: record what a web view already taught us

jerimiah797:
Answers "what can we save so rediscovery is minimised". Follows #93.

## The measurement

Metatext's Account Settings, an out-of-process `SFSafariViewController`:

| | probes | time | elements |
|---|---|---|---|
| cold | 17 | ~3.5s | 13 |
| with the origin recorded | **1** | **~0.2s** | 13 |

Out-of-process views need this most: they carry their own chrome at *both* ends,
none of it in the native tree, so geometry cannot even propose an origin and the
sweep is the only route.

## A recorded origin is a hypothesis, never a fact

It depends on device size, iOS version and text size. So it enters as the
**first candidate** and is confirmed by one probe exactly like any other; if it
no longer holds it is discarded and the ordinary search proceeds. That is the
whole reason it is safe to write down — trusting it would mean a confident tap
on whatever now occupies that pixel, which is the failure this design keeps
refusing to allow. Both directions are tested.

## Matched by URL, not by screen

The pages a hint could help locate are largely the ones with **no native
identity** — the tree reports a single element for a settings page. Requiring
the screen to be identified first would rule out every case worth optimising.
For the same reason `web_content:` is parsed even when a screen has
`landmarks: []`, which is precisely where these screens live; a hint reachable
only through a successful landmark parse would never reach the cases it exists
for.

## A documentation correction

The guide claimed system-presented views are closed to the Web Inspector "no
matter what the app does". I wrote that in #90 by inference. It is **wrong**:

| view | Web Inspector | hit-test |
|---|---|---|
| in-process `WKWebView` | yes — if the app sets `isInspectable` | yes |
| `SFSafariViewController` | **yes, with no opt-in from the app** | yes |
| `ASWebAuthenticationSession` | **no — zero connected applications** | yes |

The old `web_content:` schema (`in_process`, `inspectable`, `page_offset`) was
speculative, written before anchoring existed. Replaced with the measured one.

## Round trip

`get_web_content` now returns what it resolved, so a discovery can be recorded
without measuring it by hand:

```json
{"page_id": 3, "url": "https://social.arctian.org/settings",
 "origin": [0, 106], "viewport": [402, 685], "strategy": "sweep"}
```

`strategy: "hint"` means the recorded value was used and confirmed; `geometry`
or `sweep` means it was rediscovered — the signal to write the new value down.

Verified end to end against the live simulator: measured cold, pasted into
`account-settings.md`, reloaded, and the next three calls came back
`strategy: hint`, 1 probe.

Suite 1590, ruff clean, new guards mutation-verified.

## Still to come, from the same design discussion

- `get_web_content` returning elements for `ASWebAuthenticationSession` via the
  probe sweep, so one tool covers all three cases.
- URL as a landmark kind, giving identity to the 7 screens that have none.

## Summary by CodeRabbit

* **New Features**
  * Improved web content discovery across supported web-view types using saved URL, process, reachability, and viewport information.
  * Web content results now include page URL, origin, viewport, and discovery method.
  * Added support for storing and reusing web-view hints per app.

* **Documentation**
  * Expanded guidance for web-view inspection, hit testing, probe strategies, and third-party URLs.
  * Updated templates with examples for supported web-view behaviors.

* **Tests**
  * Added coverage for hint validation, reuse, fallback discovery, URL matching, and lifecycle cleanup.

### DIFF
--- a/docs/app-knowledge-guide.md
+++ b/docs/app-knowledge-guide.md
@@ -246,45 +246,80 @@ This is a separate issue from identifier reliability. The label is *correct*, bu
 ### Documenting Web Views
 
 A screen built on web content looks almost empty to `get_ui_tree`: on iOS the
-accessibility tree does not descend into a `WKWebView`, and the out-of-process
-hosts are not in the app's hierarchy at all. The failure mode is a
-misdiagnosis, not an error — a near-empty tree reads as "the screen failed to
-load" or "every landmark drifted", and the agent goes looking for a problem
-that isn't there.
+accessibility tree does not descend into a `WKWebView`, and out-of-process hosts
+are not in the app's hierarchy at all — a settings page presented in
+`SFSafariViewController` reports **one** element, the Application. The failure
+mode is a misdiagnosis, not an error: a near-empty tree reads as "the screen
+failed to load" or "every landmark drifted", and the agent goes looking for a
+problem that isn't there.
 
-Record it in the screen's `web_content:` block. The facts that matter are ones
-only the source can answer, which is exactly what a knowledge base is for.
+`get_web_content` reads the page and returns its elements with real screen
+frames. What it cannot work out for itself is which process hosts the view and
+whether it can be inspected at all — so record that.
 
 ```yaml
 web_content:
-  - host: "WKWebView"
-    in_process: true
-    inspectable: "debug"
-    url: "https://joinmastodon.org/servers"
-    page_offset: [0, 100]
+  - host: "SFSafariViewController"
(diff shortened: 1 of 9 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #95, opened 2026-09-03
chore: stop teaching a field nothing reads

jerimiah797:
## The framing was wrong, and the real problem is worse

This started as "is `identify_by:` legacy support worth keeping for old users".
Checking the timeline gave a different answer:

- `identify_by:` was the only convention for **five weeks** — v0.10.0 (Mar 21)
  to v0.11.0-rc1 (Mar 29). Landmarks shipped in v0.11.0 on Apr 26.
- But **the templates never stopped emitting it.** So this is not a legacy
  convention surviving in old knowledge bases. It is one *still being created* —
  every knowledge base written since April, including this week's, carries a
  field the loader has never read.

The loader's handling is **ten lines**. The cost was in the templates, the stub,
the glossary and ~23 doc mentions, all still teaching it.

## Removed

- `identify_by:` from the screen template, the stub and the glossary.
- The `quern-landmark-migration` skill. Five weeks of exposure, and the rename
  is mechanical — the loader's diagnostic already returns the entries needed to
  do it.
- ~23 doc mentions down to 5, each now load-bearing.

## Kept, deliberately

The ten lines that recognise the old field, reframed from a migration path to a
**diagnostic**. Without them a file with only `identify_by:` reports
`no_landmarks`, which tells the reader nothing. With them the original entries
come back in the response, which is enough to do the rename from it directly.

Existing knowledge bases are unaffected — nothing about how they load changes.

## A gap this exposed

Skills are symlinked into `~/.claude/skills/`, and `_install_skills` only ever
*added* them. So retiring any skill leaves a link to a directory that no longer
exists, on every machine that ever ran setup — this one included, which is how I
noticed. Setup now prunes links into its own skills directory whose target is
gone, and leaves everything else alone: a broken link to somewhere else is not
ours to remove, and a real directory of the same name is the user's own skill.
Both cases tested and mutation-verified.

## One thing left alone

`alerts/_template.md` keeps `identify_by:`. Alerts are not scanned by the
landmark loader at all, so there it is the current schema rather than a
superseded one — it is the only recognition info those files carry. The shared
name across two schemas is a trap, but renaming it is a separate decision.

## Verified against a real knowledge base

Stripped all 39 Metatext screen files (alerts restored — see above) and
reloaded: **29 screens, 0 collisions**, unchanged. The 8 skips moved from
`legacy_format` to `no_landmarks`, which is now the honest reason: they have
`landmarks: []` and nothing else.

Suite 1597, ruff clean.

## Summary by CodeRabbit

- **Documentation**
  - Clarified that screen matching uses `landmarks`, with guidance for legacy files, malformed entries, troubleshooting, and migration.
  - Updated templates and glossary content to reflect landmark-based screen recognition.
  - Clarified that legacy `identify_by` entries are preserved as diagnostic information rather than evaluated automatically.

- **Bug Fixes**
  - Improved skill setup handling for relative stale links and unlink failures so remaining skills continue installing.

- **Removed**
  - Retired the landmark migration skill and removed legacy `identify_by` guidance from new templates.


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #96, opened 2026-09-03
feat: reach web content no process will describe

jerimiah797:
The last of the three web-view cases, and the one that matters most in practice.

## The gap

An `ASWebAuthenticationSession` is presented with **no connected application at
all** — the Web Inspector reports zero — while the accessibility tree reports
**one** element, the Application. So the OAuth screen that nearly every app has
was the one screen `get_web_content` could say nothing about.

Hit-testing reaches it. When the Inspector returns nothing, `get_web_content`
now sweeps the screen instead, aimed by Vision text recognition. Those elements
are already in screen coordinates, so there is no page space to anchor; what is
lost is the DOM, so no ids, tags or hrefs are reported rather than invented.
`route` says which one answered.

The sweep is the route of last resort, not a supplement — a screenshot, text
recognition and a dozen or so probes — so it runs only when the Inspector
produced nothing. Both directions are mutation-tested.

## The proof

The Metatext OAuth sign-in against a self-hosted GoToSocial instance, **driven
entirely through the HTTP API** — no `sim_bridge` import, no `describe_point`,
only endpoints an agent has:

```
read: route=probe 13 elements (3.4s)
Email    tap=ok typed (5.4s)
Password tap=ok typed (9.8s)
Sign in  tap=ok (13.4s)
after: [... 'Authorize app', 'arctian_test01', 'Allow' ...]
total 22.7s
```

then `tap_element(label="Allow")` → `ok`, `source: web-probe`, back on the
timeline. Before this, that flow could only be driven by a Python script
importing Quern's internals.

## A bug this exposed

`clear_text`'s docstring said it clears **the focused field**. It actually
clears **the first field that has a value** — so on a sign-in form, clearing
before typing the password finds the *email* field and empties that. Measured
live: the email arrived as `arctian_test` and the page refused it.

It now takes a `label` or `identifier`, and the description says plainly what it
does without one. Focus genuinely cannot be read — the accessibility tree does
not report it — so naming the field is the only correct answer, and the fallback
is documented rather than disguised.

## Worth knowing for scripting

The probe route costs ~2s per read, and an overlay is dropped by every tap. A
script that re-reads before *every* action took long enough that GoToSocial's
OAuth session expired mid-flow (`redirect_uri not found in session`) — twice.
Reading only when the overlay has actually been dropped brought the same flow
to 22.7s and it succeeded. That pacing note is now in the tool description.

Suite 1610, ruff clean, MCP builds. New guards individually mutation-verified —
including one that initially passed for the wrong reason, exercising the
verification helper rather than the `tap_element` gate that actually decides.

## Summary by CodeRabbit

* **New Features**
  * Improved web content discovery across supported in-app browsers using Web Inspector with a text-recognition fallback.
  * Web content results now identify their discovery method and support more reliable tapping.
  * Added targeted text-field clearing by label or identifier, while preserving default first-field behavior.
  * Added support for clearing fields in SafariViewController and ASWebAuthenticationSession contexts.

* **Documentation**
  * Expanded guidance on discovery methods, route selection, compatibility, and performance.

### DIFF
--- a/docs/app-knowledge-guide.md
+++ b/docs/app-knowledge-guide.md
@@ -279,11 +279,13 @@ not follow the rule you would guess:
 | `SFSafariViewController` | **yes, with no opt-in from the app** | yes |
 | `ASWebAuthenticationSession` | **no — reports no connected application at all** | yes |
 
-`get_web_content` tries both, Inspector first, and reports which answered in
-`route`. Recording `reachable_by` still pays: a screen known to be
-`hit_test`-only skips a failing Inspector attempt, and the probe route costs a
-screenshot, text recognition and a dozen or so hit-tests — around 2s, against
-~0.2s for an Inspector page with a recorded origin.
+`get_web_content` always asks the Inspector first and falls back to probing only
+when it returns nothing, reporting which answered in `route`. `reachable_by`
+does not change that order — it tells a reader what to expect, and what a result
+costs: the probe route needs a screenshot, text recognition and a dozen or so
+hit-tests, around 2s, against ~0.2s for an Inspector page with a recorded
+origin. A screen recorded as `hit_test` only will always pay the failed
+Inspector attempt first.
 
 The middle row is the surprising one. The app has no handle on a
 `SFSafariViewController`, yet WebKit still publishes it for inspection under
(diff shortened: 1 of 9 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #101, opened 2026-09-04
feat: wait for the screen to settle instead of sleeping

### LINKED ISSUE #100: No way to wait for the screen to settle; interacting early fails silently
## What happens

There is no way to ask "has the screen finished drawing?", and interacting before it has loses the interaction without reporting anything. Web views are the worst case, because the accessibility tree cannot see into them at all, but the problem is general.

Measured against a web `<input>` in an out-of-process `SFSafariViewController` (GoToSocial's `/settings` page). Only the delay between presenting the view and tapping the field varies; each run then types one character and reads the value back:

| approach | typed character landed | cost |
|---|---|---|
| `sleep(0.2)` | **1/5** | 0.2s |
| `sleep(1.0)` | 5/5 | 1.0s |
| **wait until the screen settles** | **5/5** | **0.65–0.69s** |

`tap` and `type_text` returned success on every run, including the four that did nothing.

The settle-based runs used **no sleeps anywhere** — not before the tap, not between navigation steps. It is both more reliable than the short sleep and faster than the safe one, because it waits for the actual condition instead of a guess about it.

## Why a sleep can never be right here

The page above was **cached** — the sheet was closed and reopened, not loaded cold. A first load, or a slow network, moves the window arbitrarily far out. Any constant is simultaneously too long for the fast case and too short for the slow one.

Quern already has four of these, and each comment names the condition it is standing in for:

| | |
|---|---|
| `controller_ui.py:503` | *"a swipe can leave the container rubber-banding"* → `sleep(0.3)` |
| `controller_ui.py:1393` | *"let the scroll settle before clicking"* → `sleep(0.4)` |
| `controller_ui.py:1588` | *"wait longer for animation to complete"* → `sleep(0.3)` |
| `controller_ui.py:1558` | stability check: `sleep(0.1)`, then re-fetch and compare element frames |

The last is the same idea already, but it compares *accessibility frames*, so it is blind to web content, to anything still painting inside a settled frame, and it costs a full tree fetch per poll.

## Proposal: settle by comparing screenshots

Capture, downscale, compare consecutive frames, and stop when the change falls below a threshold for two consecutive comparisons — bounded by a timeout.

It needs no privileged channel, which is what makes it general: it works for an `ASWebAuthenticationSession` where the Web Inspector reports **no connected application at all**, for native transitions, and on Android.

Measured, on this machine against a booted simulator:

- **capture + downscale to quarter size: ~208ms; capture + compare: ~134ms.** The capture dominates; the comparison is nearly free.
- **Idle screen noise is 0.00017** of pixels changed frame to frame — not zero. A threshold of **0.001** clears it with room to spare, and is not a guess: it comes from that measurement.
- **A blinking caret does not defeat it.** With a focused text field, 41 of 45 frame pairs were still under threshold (the caret peaks at 1.8% of pixels). It delays settling by a blink at worst.
- **Genuinely animated content never settles** — a spinner, a playing video, a carousel. That is what the timeout is for, and why the result must distinguish "settled" from "timed out" rather than silently proceeding.

Sketch:

```python
async def wait_settled(udid, *, interval=0.25, stable_needed=2,
                       epsilon=0.001, timeout=20.0) -> SettleResult
```

## Where it should be used

1. **A tool of its own**, so an agent can wait explicitly.
2. **Inside the interaction paths**, replacing the four sleeps above.
3. **Before the pre-tap probe** for web elements, which currently verifies against whatever happens to be drawn at that instant.

## Related

- #99 — `type_text` reports success regardless, which is what makes this silent.
- #98 — `clear_text` removes an arbitrary amount of a web field; the other half of why filling a web form is guesswork today.

## Not this

An earlier version of this issue proposed waiting on `document.readyState`. That is worse: it only exists for pages the Web Inspector can see (so not auth sessions), and it is the wrong signal anyway — a page can be interactive before `complete`, and `complete` before layout has settled or images have painted.

jerimiah797:
Closes #100.

## The measurement that motivated it

A web form where typing is silently lost if it arrives early, varying only the
wait:

| approach | keystroke landed | cost |
|---|---|---|
| `sleep(0.2)` | **1/5** | 0.2s |
| `sleep(1.0)` | 5/5 | 1.0s |
| **wait until settled** | **5/5** | **0.65–0.69s** |

Faster than the sleep that was reliable; reliable unlike the sleep that was
fast. The page in that test was **cached** — a cold load moves the window
arbitrarily, which is the whole argument against a constant.

## The one that justifies the change to existing code

A real scroll on a populated timeline takes **~1100ms** to settle (8 frames,
±15ms across five runs). **The sleep it replaced was 300ms.**

That sleep's own comment described the bug it was there to prevent — an
over-scroll bounce read as a resting position — so the constant was not merely
imprecise, it was 3.6× too short to do its job. This is a correctness fix, not
a speed one.

## Nothing here is a guess

- **Threshold 0.001** of pixels: an idle simulator measures **0.00017** frame to
  frame. A blinking caret peaks at 1.8% and so exceeds it, delaying settling by
  one blink rather than defeating it (41/45 pairs still under, measured on a
  focused field).
- **Two consecutive unchanged comparisons**, because a frame taken
  mid-animation can match its predecessor by coincidence at the extreme of an
  ease curve.
- **No delay between captures.** A capture already costs ~114ms; adding an
  interval would reintroduce the constant this exists to remove. Raw capture,
  not the processed screenshot — that saves ~62ms per frame that would be spent
  resizing an image this decodes itself anyway.

## What it deliberately does not do

**It answers "has drawing stopped", not "has content arrived."** A blank page
part way through a slow load is perfectly still. Measured against a request to a
non-routable address: Safari showed a white screen with its progress bar
stalled, and this reported `settled: true` after **1.6s** with zero change
between frames. **A longer timeout does not help, because nothing is moving.**
That is stated in the tool description, the module docstring and the endpoint —
it is the sharpest edge here and the easiest to misread.

## Pillow, not Vision — having measured both

| | pixel difference | Vision feature print |
|---|---|---|
| cost per frame | ~17ms | **~98ms** |
| caret-sized change (2×30) | 0.00009 | 0.0386 |
| one line of text (400×40) | 0.00599 | 0.0923 |
| usable threshold window | **66×** | **2.4×** |

I expected feature prints to be *less* sensitive to small changes. They are
more. But that is the wrong direction here: a caret is a tiny change that never
stops, so it must be tolerated or a focused field never settles — and Vision
leaves only a 2.4× margin between tolerating one and catching a line of text.
It would also make this macOS-only, where `raw_screenshot` already handles adb.

(Incidental: pyobjc does not expose `computeDistance:`'s out-parameter without
`registerMetaDataForSelector`, and silently returns `(True, None)` otherwise.)

## Scope

Two of the four sleeps became waits. The Android one is left alone — adb's
capture time is unmeasured here, so swapping it blind could make that path
slower rather than more reliable. The 0.1s stability pause is left too, being
shorter than three captures cost.

13 new tests, each mutation-verified. Note one of those mutations initially
**hung the suite** rather than failing it: `asyncio.wait_for` cannot cancel a
poll loop whose await never suspends, so the brake had to go in the capture
itself, which counts its own calls and raises. Suite 1648, ruff clean, MCP
builds, README/api-reference guard-tested.

## Summary by CodeRabbit

* **New Features**
  * Added a tool that waits until the device screen becomes stable before continuing.
  * Added an API endpoint with optional device selection and timeout settings.
  * iOS screen interactions now wait for visual stability instead of using a fixed delay.

* **Documentation**
  * Updated the README and API reference with the new tool and revised tool count.

* **Tests**
  * Added coverage for stable screens, ongoing motion, timeouts, capture failures, and image-difference thresholds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,7 +280,7 @@ quern tunneld <cmd>          # Manage the tunneld LaunchDaemon (install/uninstal
 
 ## MCP Tools
 
-108 tools available via MCP. All tools are lazy-loaded and won't hog your context just by connecting the MCP. They are lightweight API wrappers and are easy for the Agent to use.
+109 tools available via MCP. All tools are lazy-loaded and won't hog your context just by connecting the MCP. They are lightweight API wrappers and are easy for the Agent to use.
 
 | Category | Tools |
 |----------|-------|
@@ -291,7 +291,7 @@ quern tunneld <cmd>          # Manage the tunneld LaunchDaemon (install/uninstal
 | System Proxy | `configure_system_proxy`, `unconfigure_system_proxy` |
 | Intercept & Mock | `set_intercept`, `clear_intercept`, `list_held_flows`, `release_flow`, `replay_flow`, `set_mock`, `list_mocks`, `update_mock`, `clear_mocks` |
 | Device | `list_devices`, `boot_device`, `shutdown_device`, `erase_device`, `install_app`, `launch_app`, `terminate_app`, `uninstall_app`, `list_apps`, `build_and_install` |
-| UI | `get_ui_tree`, `get_element_state`, `wait_for_element`, `get_screen_summary`, `tap`, `tap_element`, `swipe`, `scroll_to_element`, `type_text`, `clear_text`, `press_button`, `get_web_content` |
+| UI | `get_ui_tree`, `get_element_state`, `wait_for_element`, `get_screen_summary`, `tap`, `tap_element`, `swipe`, `scroll_to_element`, `type_text`, `clear_text`, `press_button`, `get_web_content`, `wait_for_settle` |
 | Screenshots | `take_screenshot`, `take_annotated_screenshot`, `start_screenshot_timeline`, `stop_screenshot_timeline`, `get_screenshot_timeline` |
 | Device Config | `set_location`, `open_url`, `grant_permission`, `set_locale`, `set_hardware_keyboard`, `set_font_scale`, `set_display_density` |
 | App State | `save_app_state`, `restore_app_state`, `list_app_states`, `delete_app_state` |
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -112,6 +112,7 @@ The key lives at `~/.quern/api-key`; the server's URL and port are in `~/.quern/
 | `swipe` | POST | `/api/v1/device/ui/swipe` | Swipe gesture |
 | `scroll_to_element` | POST | `/api/v1/device/ui/scroll-to-element` | Scroll a container until the target is in view, without tapping it |
 | `get_web_content` | POST | `/api/v1/device/ui/web-content` | Read WKWebView content the accessibility tree cannot see (iOS simulator only) |
+| `wait_for_settle` | POST | `/api/v1/device/ui/wait-settled` | Wait until the screen stops changing, by comparing successive screenshots |
 | `type_text` | POST | `/api/v1/device/ui/type` | Type text |
 | `clear_text` | POST | `/api/v1/device/ui/clear` | Clear text field |
 | `press_button` | POST | `/api/v1/device/ui/press` | Press hardware button |
--- a/mcp/src/tools/device-ui.ts
+++ b/mcp/src/tools/device-ui.ts
@@ -508,6 +508,48 @@ Label matching modes (mutually exclusive — use only one):
     }
   });
 
+  server.registerTool("wait_for_settle", {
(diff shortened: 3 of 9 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #103, opened 2026-09-04
fix: actually clear a web text field

### LINKED ISSUE #98: clear_text removes an arbitrary amount of a web text field, and reports ok either way
## What happens

`clear_text` empties a **native** field completely and a **web** field partially, by an amount that varies from one character to nearly all of it. Every call returns `{"status": "ok"}`.

Native `TextArea` (Metatext's compose box, `composition.text`) — correct:

```
typed 60 chars -> value length 60
after clear #1 -> 0
```

Web `<input>` inside an out-of-process `SFSafariViewController` (GoToSocial's `/settings` page, the "Instance" field) — one character:

```
before clear: 26  'https://social.arctian.org'
after  clear: 25  'https://social.arctianorg'
```

Web `<input>` in the sign-in form of the same site, with a longer value — 137 characters:

```
before clear: 168
after  clear: 31
```

## Why

`SimBridgeBackend.select_all_and_delete` (`server/device/sim_bridge.py:550`) selects by **triple-tapping**, then sends a single backspace:

```python
# Triple-tap to select all
for _ in range(3):
    await self._send({"cmd": "tap", "udid": udid, "x": x, "y": y})
    await asyncio.sleep(0.05)
await asyncio.sleep(0.1)
# Send backspace (HID usage 0x2A on page 7)
await self._send({"cmd": "type", "udid": udid, "text": "\x08"})
```

A native text view honours the triple-tap as select-all, so the one backspace clears everything. A web input evidently does not do so consistently — and whatever the selection turns out to be, exactly one backspace follows it. When nothing is selected, that deletes a single character.

The 50ms spacing may be part of it; the taps have to register as one triple-click rather than three separate ones. I did not isolate that, so treat the mechanism below the triple-tap as unconfirmed. The measurements above are direct.

## Why it matters

A caller that clears before typing gets the remains of the old value stitched onto the new one, and nothing in the response says so.

Measured consequence, during an OAuth sign-in against a self-hosted GoToSocial instance: clearing the email field left `arctian_test`, the typed address was appended to that, and the page rejected the result. The visible failure was *"Enter an email address"* on a login form — several steps away from the call that actually went wrong. Retrying made it worse, since each attempt appended another copy (26 → 90 → 116 → 142 → 168 characters).

Web fields are exactly where this bites, because they are also where a caller is most likely to be filling a form it cannot see natively.

## Notes

- `get_web_content`'s probe route reports a web field's `value`, so a caller *can* verify a clear landed — but has to know to.
- The `clear_text` tool description now warns that a long value may need clearing more than once. That is a workaround, not a fix: the failure is silent and the required number of calls is not knowable in advance.

jerimiah797:
Closes #98.

## The bug

`clear_text` selected by **triple-tapping**. A native text view honours that; a
web input generally does not, and the single backspace that followed then
removed **one character** — while the call reported `ok`.

Measured inside an `SFSafariViewController`:

| field | before → after, one call |
|---|---|
| native `TextArea` | 60 → **0** ✓ |
| web `<input>` | 26 → **25** (at every tap spacing tried: 0.05s … 0.25s) |
| web `<input>`, longer value | 168 → **31** (a partial selection happened to take) |

The consequence during an OAuth sign-in was an email arriving as
`arctian_test` and the page refusing it — several steps away from the call that
actually went wrong.

## The fix, in tiers

**Native fields are untouched.** The triple-tap works there. Re-reading to
confirm would cost a full tree walk on *every* clear to learn something already
known, so the element's source decides which kind of field this is.

**Web fields, first choice — the Web Inspector.** `6ms`, against ~3.5s of
keystrokes for a 55-character value.

It deliberately does **not** assign through `el.value`. React and friends
override that setter on the element instance, so the DOM would empty while the
framework's own state kept the old value — a field that looks cleared and is
not, which is the same bug in a new costume. It goes through the prototype's
setter and fires `input` and `change`. Only the *focused* element is touched, so
asking every page cannot wipe a field on a page the caller was not typing into.

**Web fields, fallback — backspaces.** One per character (~63ms each). Depends
on nothing but the backspace key, so it works where the Inspector sees nothing
at all: an `ASWebAuthenticationSession` reports **no connected application**,
and that is the OAuth form every app has. The caret is moved to the trailing
edge first, since backspace only removes what is behind it.

Measured end to end: web field 60 → 0 in **1.23s** (was 7.65s via keystrokes
alone), native 61 → 0 in **0.77s**.

## A bug this exposed

`connected_applications` answered from a cache built during the connect
handshake. On a long-lived connection an app that launched **later was
invisible** — the server's own connection could not see the `SafariViewService`
hosting the page directly in front of it, so `get_web_content` fell back to
probing while reporting nothing wrong. That is why the fast path kept losing to
keystrokes here, and it was silently degrading web-content reads generally.

The channel is now drained briefly before answering. Note the first version of
that fix made things worse: `connected_applications` had never touched the
socket, so a dead connection began raising where callers expected a list, and
that surfaced as a 500. It answers from cache on a failed drain instead.

## Verification

5 new tests, each mutation-verified — native-skips-the-extra-work,
keystroke-fallback, caret-placement, and inspector-preference all fail when
their guard is removed. Suite 1656, ruff clean, MCP builds.

## Note

`clear_text` on a web field still requires a prior `get_web_content`, since the
field is only addressable through the overlay. Without one it correctly returns
"No text field found to clear" rather than clearing something else.

## Summary by CodeRabbit

* **New Features**
  * Improved clearing of text in web-based input fields using Web Inspector support when available.
  * Added fallback keystroke deletion when Inspector clearing is unavailable or incomplete.
  * Added refreshed application information to improve clearing reliability.

* **Bug Fixes**
  * Web overlay fields now retain their displayed values.
  * Clearing is prevented from partially deleting oversized web fields.
  * Clearing results from unrelated simulator applications are no longer accepted.
  * Native fields avoid unnecessary deletion steps after clearing.

### DIFF
--- a/mcp/src/tools/device-ui.ts
+++ b/mcp/src/tools/device-ui.ts
@@ -705,7 +705,7 @@ Label matching modes (mutually exclusive — use only one):
   });
 
   server.registerTool("clear_text", {
-    description: `Clear a text field (select-all + delete), before type_text when the field already holds content. Pass label or identifier to say WHICH field: without one this picks the first field that has a value, which is not the field you just tapped -- on a sign-in form, clearing before typing the password finds the email field and empties that instead. Focus cannot be detected; the accessibility tree does not report it. Select-all takes a line at a time, so a long or wrapped value may need clearing more than once; check the value and repeat. Secure text fields (passwords) may not support select-all at all.`,
+    description: `Clear a text field (select-all + delete), before type_text when the field already holds content. Pass label or identifier to say WHICH field: without one this picks the first field that has a value, which is not the field you just tapped -- on a sign-in form, clearing before typing the password finds the email field and empties that instead. Focus cannot be detected; the accessibility tree does not report it. A web field is cleared through the Web Inspector when the page is visible to it (a few milliseconds), and by keystroke otherwise -- about 63ms per character, so a long value takes seconds. Either way the field ends empty. Secure text fields (passwords) may not support select-all at all.`,
     inputSchema: strictParams({
       label: z
         .string()
--- a/server/device/controller_ui.py
+++ b/server/device/controller_ui.py
@@ -2174,25 +2174,24 @@ class DeviceControllerUI:
 
         cx = target.frame["x"] + target.frame["width"] / 2
         cy = target.frame["y"] + target.frame["height"] / 2
-
         backend = self._ui_backend(resolved)
-        await backend.select_all_and_delete(
-            resolved, x=cx, y=cy, element_type=target.type,
-        )
 
-        # Native fields are done: the triple-tap selects the whole value and the
-        # backspace clears it, measured at 60 characters to 0 in one call.
-        #
-        # A web input usually does not honour the triple-tap, and the backspace
-        # then removes a single character -- measured on a form inside an
-        # SFSafariViewController, one call took 26 characters to 25, so the
-        # field looks cleared without being cleared. Only those need more work,
-        # and the source says which is which. Deciding by re-reading the field
-        # instead would cost a full tree walk on every clear, native ones
-        # included, to learn something already known.
+        # The triple-tap selects the whole value in a native text view --
+        # measured at 60 characters to 0 in one call -- and is all that is
+        # needed there.
         if (target.extra_attrs or {}).get("source") not in ("web-inspector", "web-probe"):
+            await backend.select_all_and_delete(
+                resolved, x=cx, y=cy, element_type=target.type,
+            )
             self._invalidate_ui_cache(resolved)
             return resolved
+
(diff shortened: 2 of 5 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #104, opened 2026-09-04
fix: let type_text prove the text arrived

### LINKED ISSUE #99: type_text reports success when there is nowhere for the text to go
## What happens

`type_text` returns `{"status": "ok"}` whether or not the text reached a field. On a screen with **no text field at all**:

```
GET  /api/v1/device/ui         → 14 elements, 0 text fields

POST /api/v1/device/ui/type    {"udid":"<sim>","text":"this goes nowhere at all"}
→ HTTP 200  {"status":"ok","udid":"<sim>"}
```

`clear_text` on the very same screen gets this right, which is what makes the contrast awkward:

```
POST /api/v1/device/ui/clear   {"udid":"<sim>"}
→ HTTP 500  {"detail":"[idb] No text field found to clear"}
```

## Why

`DeviceControllerUI.type_text` forwards to the backend and returns:

```python
async def type_text(self, text: str, udid: str | None = None) -> str:
    """Type text into focused field. Returns the resolved udid."""
    resolved = await self.resolve_udid(udid)
    await self._ui_backend(resolved).type_text(resolved, text)
    self._invalidate_ui_cache(resolved)
    return resolved
```

Nothing establishes that a field has focus beforehand, and nothing checks that a value changed afterwards. The docstring says "into focused field", but focus is neither verified nor verifiable — the accessibility tree does not report it. So the call describes an intent, and reports on having made the attempt.

## Why it matters

This is the [#78](https://github.com/quern-dev/quern/issues/78) pattern again: **a well-formed success for something that didn't happen.** An agent types a value, gets `ok`, and asserts on a screen that never changed — so the failure surfaces later, somewhere unrelated.

Observed repeatedly while automating a login flow. A tap that lands on a field but does not take focus (which happens on a freshly presented screen — I saw it on Metatext's `add-identity.url-field` across several runs) leaves the subsequent `type_text` writing into nothing, and both calls report success. Every script I wrote against that flow eventually grew the same shape:

```python
for attempt in range(4):
    tap(field); type_text(value)
    if read_back(field) == value:
        break          # only way to know it worked
```

That loop is the workaround every caller has to reinvent.

## What would help

Some evidence, rather than none. In rough order of usefulness:

1. **Verify by read-back** — compare the target field's value after typing, and report a mismatch. Costs one UI read; the caller is going to do it anyway.
2. **Accept a target** — `label`/`identifier` like `clear_text` now takes, so the call knows which field it means and can tap it first rather than relying on ambient focus.
3. **At minimum, fail when there is no text field on screen**, matching `clear_text`. That alone would have caught the case above.

I have not attempted a fix; filing so the behaviour is recorded. Note that a focus failure has not been reproduced deterministically — 12/12 attempts landed on a native field in a controlled loop — but the reporting defect above is structural and reproduces every time.

jerimiah797:
Closes #99.

## The bug

`type_text` returned `ok` whether or not the keystrokes reached a field. The
sharpest demonstration is that `clear_text`, in the same module, already got
this right:

```
screen with 0 text fields
POST /device/ui/type   → 200 {"status":"ok"}          ← typed into nothing
POST /device/ui/clear  → 500 "No text field found"    ← correct
```

And a tap that failed to take focus looked identical to one that worked — which
is how an OAuth sign-in submitted an address the page rejected several steps
later, with every call reporting success.

## The fix

Name the field and the call proves itself: located, tapped, typed into, read
back. A value that did not move is an error. A field that is not there is
refused, matching `clear_text`.

```
POST /device/ui/type {"text":"…","identifier":"composition.text"}
  → 500  "No text field matching 'composition.text' to type into"   (nothing there)
  → 200  {"status":"ok","verified":true}                            (it landed)
POST /device/ui/type {"text":"…"}
  → 200  {"status":"ok","verified":false}                           (untargeted)
```

## Why verification is opt-in

It costs a **fresh** tree read — ~1.8s natively, ~2s for web content, because
the overlay still holds the value from *before* the keystrokes, so consulting it
would only confirm what was already believed. Making that the default would tax
every keystroke in every flow to catch a case most of them never hit.

Untargeted typing therefore keeps its behaviour and simply stops implying more
than it knows: `verified: false`. Nothing in the accessibility tree reports which
element has focus, so that is the honest answer rather than a guess.

## One detail worth naming

A secure field is verified **by length**, not content: it reports dots, so the
typed text is never visible in the value and a containment check would fail
every time. An ordinary field still checks content, so text landing in some
*other* field cannot pass.

Suite 1664, ruff clean, MCP builds. Five new guards, each mutation-verified.

## Summary by CodeRabbit

* **New Features**
  * Added optional label and identifier targeting when entering text.
  * Label and identifier selectors can be combined to locate a specific field.
  * Targeted text entry now confirms the expected value was entered and reports verification status.
  * Secure fields can be verified by text length.
  * Existing untargeted text entry remains supported, with verification reported as unavailable.

* **Bug Fixes**
  * Improved handling when fields cannot be found or entered text does not match the expected value.

### DIFF
--- a/mcp/src/tools/device-ui.ts
+++ b/mcp/src/tools/device-ui.ts
@@ -653,10 +653,12 @@ Label matching modes (mutually exclusive — use only one):
     inputSchema: strictParams({
       label: z
         .string()
+        .min(1)
         .optional()
         .describe("Field to type into, by visible text. Enables verification."),
       identifier: z
         .string()
+        .min(1)
         .optional()
         .describe("Field to type into, by accessibility identifier. Enables verification."),
       text: z.string().describe("Text to type"),
--- a/mcp/src/tools/device-ui.ts
+++ b/mcp/src/tools/device-ui.ts
@@ -649,8 +649,16 @@ Label matching modes (mutually exclusive — use only one):
   });
 
   server.registerTool("type_text", {
-    description: `Type text into the currently focused input field. On iOS simulators this attaches the simulated hardware keyboard before typing (required for shifted characters to retain their modifier), which hides the software keyboard — use set_hardware_keyboard with enabled=false afterward if a later step expects the software keyboard to be visible.`,
+    description: `Type text into an input field. PASS label OR identifier whenever you can: the field is then located, tapped, typed into and read back, so the call fails loudly if the text did not arrive. Without one the text goes wherever the keyboard happens to be pointed and nothing can check it -- on a screen with no text field at all this still returns ok, having typed into nothing, and a tap that failed to take focus looks identical to one that worked. The response reports verified: true|false so you can tell which you got. Verification costs a fresh UI read (~2s), which is why it is opt-in. On iOS simulators this attaches the simulated hardware keyboard before typing (required for shifted characters to retain their modifier), which hides the software keyboard — use set_hardware_keyboard with enabled=false afterward if a later step expects the software keyboard to be visible.`,
     inputSchema: strictParams({
+      label: z
+        .string()
+        .optional()
+        .describe("Field to type into, by visible text. Enables verification."),
+      identifier: z
+        .string()
+        .optional()
+        .describe("Field to type into, by accessibility identifier. Enables verification."),
       text: z.string().describe("Text to type"),
       udid: z
         .string()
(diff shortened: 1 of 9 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #107, opened 2026-09-05
Report and optionally apply external tool upgrades

jerimiah797:
> **Stacked on #106.** Base is `deps/floors-and-eager-upgrades`, so this diff
> shows only its own changes. Merge #106 first.

## The gap

`quern update` kept its own venv current and said nothing about anything else.
That is not a small gap — `pymobiledevice3` is **two installs** on a typical
machine, and the eager venv upgrade in #106 *widened* the split rather than
closing it:

```
pymobiledevice3 (library)  11.3.1  venv   ← moved by the eager upgrade
pymobiledevice3 (cli)       9.15.1  pipx   ← serves tunneld; untouched
```

`tool_versions` answered *what is installed*. This adds `tool_updates`, which
decides *what to do about it* — a different question with a different shape.

## Three findings that set the design

**The floor is not the driver.** Every floor quern declares is a `>=` and none
is load-bearing, so a floor-driven updater sits silent while tools age
indefinitely. Latest drives the offer; a floor only escalates it into a
requirement.

`CLI_FLOORS` therefore ships **empty**, and a test keeps it that way. That is
the finding, not an omission: no CLI version boundary has ever been tested
against quern, and inventing one here would repeat exactly what #106 was written
to stop. Adding an entry has to be a deliberate act with a comment saying what
breaks below it and where that was measured.

**How a tool updates is a property of its `source`, not its name.** `adb` is
left alone because it is `android-sdk` and `node` because it is `fnm` — there is
no name list. The same tool installed from brew on another machine is offered
normally, and a test pins that. An earlier sketch keyed the exclusions off names
and would have been wrong on the next machine.

**Arriving as a dependency is a reason to leave a tool alone** — upgrading it
directly can be undone by whatever pulled it in. Unless it is below a floor,
where staying put is not an option either.

## A correctness bug fixed on the way

`brew_provenance` was fetched for **`libimobiledevice` alone, hardcoded**,
because that was the only brew install on the machine it was written on.
`mitmproxy`, `adb` and `pymobiledevice3` are all brew-installable, so on a
machine that installed them that way, `requested` / `required_by` were absent —
and `requested is False` is exactly what suppresses an unsafe upgrade offer. The
*safety* check was silently missing for every brew tool but one.

Now fetched for every brew site, concurrently (each costs a `brew info` plus a
`brew uses`), with `return_exceptions=True` so one unreadable formula cannot
cost the provenance of the others.

## A failed lookup is not a clean bill of health

`brew_outdated` returns `None` for "could not ask" and `{}` for "asked, nothing
outdated". The first version collapsed them, which reports **every formula as
current on a machine with no brew installed**.

Worth flagging for review: mutation testing showed my first tests for this were
inadequate — they injected a fake `brew` returning `None`, so they exercised the
planner and never `brew_outdated` itself. A regression to `{}` on failure passed
all 37 tests. Five tests were added against the real function.

## Surfaces

| command | behaviour |
|---|---|
| `quern update` | reports stale tools with exact commands, on **every** run — including the already-up-to-date path, since external tools age independently of quern |
| `quern update --tools` | applies them |
| `quern doctor` | shows **every** install site including healthy ones, with each brew formula's dependents |
| `quern doctor --fix` | unchanged scope (venv only), but now *says* it will not upgrade external tools |

Reporting is the default throughout: a `pipx` or `brew` upgrade changes state
for every consumer on the machine, so doing it as a side effect of updating
quern is not a decision to make on the caller's behalf.

`format_report` is deliberately separate from `format_offer`. The offer answers
"what should I run" and hides everything healthy; the report answers "why does
this machine differ from that one", where hiding the healthy entries is the
failure — two machines comparing only their problems agree they have none while
running three-major-apart copies.

The `--fix` disclaimer exists because `--fix: nothing to do` printed directly
above a tool marked `↑` reads as "and nothing to do about that either".

## Verification

- **1776 passed**, ruff clean.
- Live on this machine: correctly identifies the one stale tool, excludes `adb`
  and `node` by source, and shows `libimobiledevice`'s dependents
  (`ideviceinstaller`, `ios-webkit-debug-proxy`) on a tool needing no action.
- Mutation-tested: failed lookup collapsing to "up to date", `brew_outdated`
  returning `{}` on failure, dependency exemption removed, reporting silently
  applying, and the up-to-date path skipping the check — each caught.

## Summary by CodeRabbit

- **New Features**
  - `quern update` reports external-tool updates even when Quern is current.
  - `quern update --tools` applies eligible upgrades, including Homebrew casks.
  - `quern doctor` reports external tools without available device tools, including versions, provenance, and recommended commands.

- **Bug Fixes**
  - Corrected update commands and package mappings for externally managed tools.
  - Updates now report dependency, setup, build, and restart failures accurately.

- **Documentation**
  - Clarified update channels, virtual-environment reconciliation, `doctor --fix` behavior, and external-tool diagnostics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ quern start -f             # run in the foreground (Ctrl-C to stop)
 quern status               # check status
 quern stop                 # stop
 quern update               # update to latest release and rebuild
+quern update --tools       # ...and upgrade external tools (pipx/brew) too
 quern grant-full-perms     # allow all quern MCP tools in Claude Code without prompting
 quern mcp-install          # register quern MCP server with AI coding tools
 quern uninstall            # remove Quern and its dependencies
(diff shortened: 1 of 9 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

Issue #73, opened 2026-09-01
tunneld can wedge alive after a failed device pairing; KeepAlive cannot see it and nothing remediates


jerimiah797:
## Summary

A failed device pairing can leave `tunneld` **running but not serving**. The process stays alive, holds no listener, and never exits — so `KeepAlive => true` never fires, `launchctl` reports the job as healthy, and it stays wedged indefinitely.

Quern already detects the condition. Nothing acts on it.

## Live reproduction

Currently wedged on the work machine and being left that way deliberately, so detection schemes can be tested against a real instance rather than a synthetic one.

```
launchctl state : state = running
last exit code  : (never exited)
pid             : 949
uptime          : 15 days
tcp listeners   : 0
GET 127.0.0.1:49151 : no response
quern doctor    : ✗ tunneld
```

The daemon started ~15 days ago and served correctly for two weeks. At `Aug 31 22:37` a physical iPhone's trust prompt was denied:

```
pymobiledevice3.exceptions.UserDeniedPairingError: ('UserDeniedPairing', '00008130-...')
```

That exception stopped it serving without terminating the process. Exactly **one** occurrence in the log — it is not a crash loop. The device is no longer even connected, and the denial is not persisted anywhere; there is no state to clear, only a hung process to replace.

## Why nothing recovers it

`KeepAlive` restarts a job that **exits**. This job hangs while alive, which is invisible to launchd:

```
last exit code = (never exited)
state = running
```

From launchd's perspective the daemon is fine. It will never be restarted.

## Detection already exists — remediation is not wired to it

`is_tunneld_running()` (`tunneld.py:63`) hits the HTTP API and returns `False` right now, correctly. It has two callers:

| Caller | Behaviour when `False` |
|---|---|
| `controller.py:80` | reports `✗ tunneld` in `check_tools()` / `doctor` |
| `pmd3.py:47` | **silently falls back** to the direct usbmux path |

The second is the one with teeth. `Pmd3Backend.screenshot` picks the tunnel path or the iOS-16-and-below direct path based on that check. With tunneld wedged, an **iOS 17+ device silently takes the iOS 16 path** and fails for a reason that has nothing to do with what the user did.

Restart machinery also already exists — `tunneld.py:258-289`, `bootout` followed by `bootstrap`. So both halves are present and simply not connected.

## Suggested fix

Wire the existing check to the existing restart:

[...]

PR #108, opened 2026-09-05
Report tunneld and local-capture extension health in doctor

jerimiah797:
> **Stacked on #107** (which is stacked on #106). Base is
> `feat/external-tool-updates`, so this diff shows only its own changes. Merge
> #106 → #107 → this.

## The shared failure shape

Both services fail the same way, which is why they are one PR: **the thing quern
checks stays green while the thing that matters stops working.**

## tunneld

Reported from an HTTP probe surfaced as one boolean. That *sees* a wedge — #73
confirms doctor showed `✗` on the live wedged instance — but cannot separate it
from a daemon that was never started. Identical symptom, opposite remedy.

A failed device pairing leaves the daemon alive and not serving: it holds no
listener and never exits, so `KeepAlive` never fires and launchd reports it
healthy indefinitely. `tunneld_health()` pairs the probe with the launchd job
state, which is the signature that tells those apart:

| status | meaning |
|---|---|
| `wedged` | HTTP down, launchd says `running` — the #73 signature |
| `stopped` | HTTP down, launchd not running |
| `stale_plist` | serving, but the installed plist is outdated |
| `binary_drift` | serving from the binary the plist froze in, while quern now resolves another |
| `no_binary` / `not_installed` | prerequisites missing |

`binary_drift` is not in #73 and is invisible over HTTP, because the old binary
keeps answering perfectly well. Related to the per-user vs shared pipx path
raised under that issue's "Related".

## The mitmproxy local-capture extension

No check existed **at all**. The extension is approved once by a human in System
Settings and then upgraded underneath that approval by any ordinary dependency
update — `mitmproxy-rs` moved to 0.12.11 during the upgrade in #106. When the
shipped bundle version passes the activated one, macOS keeps running the old
extension and **local capture reports itself enabled while capturing nothing**.

Nothing else looks at this: `proxy_status` reports the mitmdump process and the
configured process list, both of which stay healthy throughout.

Verified by hand before writing the code:

```
shipped   (venv tar): org.mitmproxy.macos-redirector.network-extension  2.0/1
activated (macOS)   : org.mitmproxy.macos-redirector.network-extension (2.0/1)  [activated enabled]
```

## Both checks are unprivileged

`launchctl print` on a system-domain job and `systemextensionsctl list` are both
readable without sudo (verified on macOS 26), which is what lets read-only
diagnostics use them with no password prompt.

Reading the shipped version parses the `Info.plist` **out of the wheel's tar
without unpacking** — unpacking is what activation does, and a health check must
not have side effects. A test asserts nothing appears on disk.

## What `--fix` does and does not do

**tunneld is reported and never repaired.** Recovery is `bootout` + `bootstrap`
and belongs to #73, which has a live wedged instance being deliberately
preserved for testing; a `--fix` that silently restarted it would destroy the
only real reproduction anyone has. A test asserts doctor runs **no `launchctl`
at all**.

**`--fix` does re-run the shipped redirector app** for a stale extension, and
reports that as *handed off*, not fixed — approval is a human step in System
Settings, so claiming success would be claiming a repair that has not happened.

## Reviewer note: what is not verified

`reinstall()` has **never driven a real macOS activation prompt.** I could not
produce a stale extension without disrupting a working setup, so that path is
unit-tested only. Detection I verified by hand against `systemextensionsctl`
first. The blast radius is one `open` call on an app bundle extracted with
`filter="data"`, and it is opt-in behind `--fix`.

## Related: a correction filed on #73

That issue states *"Quern's own code already avoids `kickstart -k`."* It does
not — `install_daemon` (262) and `uninstall_daemon` (334) use `bootout`, but
`_restart_daemon` (411), reached by `./quern tunneld restart`, uses exactly
`kickstart -k`, the command the comment four hundred lines away warns *"hung
launchctl on macOS 15"*. Filed as a comment on #73, since wiring remediation to
`_restart_daemon` would inherit the hang while looking compliant. Not changed
here — that is #73's to fix.

## Verification

- **1808 passed**, ruff clean.
- Live: both report healthy on this machine, and `--fix` correctly did nothing —
  the install directory was never created.
- Mutation-tested: nested `state = active` overwriting the job state, `wedged`
  collapsing into `stopped`, a stale extension reading as healthy, a failing
  `systemextensionsctl` reading as "not registered", `--fix` restarting tunneld,
  and reporting silently reinstalling — each caught. One initially survived
  because the test passed empty stdout; it now uses a failing command that still
  prints a row.

## Summary by CodeRabbit

* **New Features**
  * Added service-health diagnostics to `quern doctor` for tunneld and the local capture extension.
  * Reports service status, installation issues, stale configurations, binary mismatches, and available remediation steps.
  * Supports optional repair of fixable capture-extension issues with `quern doctor --fix`; manual macOS approval may still be required.
  * Identifies unresponsive tunneld instances without restarting them automatically.

* **Documentation**
  * Updated the CLI reference with the expanded service-health checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,24 @@ both while one runs a 9.15.1 binary and the other an 11.3.1 one.
 tools, and says so when it finds some it cannot help with, rather than printing
 "nothing to do" above a tool it just flagged as behind.
 
+Doctor also reports **service health**, which is a different question from whether a
+tool is installed:
+
+- **tunneld** — a failed device pairing can leave the daemon alive but not serving. It
+  holds no listener and never exits, so `KeepAlive` never fires and launchd reports it
+  healthy indefinitely. Doctor separates that *wedged* state from a merely *stopped*
+  one by pairing the HTTP probe with the launchd job state, and prints the `bootout` +
+  `bootstrap` recovery. It never runs it — recovery is tracked in issue #73.
+- **local capture extension** — the mitmproxy macOS system extension is approved once
+  by a human and then upgraded underneath that approval by ordinary dependency updates.
+  When the version macOS runs falls behind the version the installed wheel ships, local
+  capture reports itself enabled while capturing nothing. Doctor compares the two, and
+  `--fix` re-runs the shipped app so macOS can activate the newer one (you still have
+  to approve it in System Settings).
+
+Neither check needs a password: `launchctl print` on a system job and
+`systemextensionsctl list` are both readable unprivileged.
+
 Setting the channel only writes `~/.quern/config.json` — nothing changes until the next `quern update`. Git-clone installs follow the `release/stable` / `release/beta` pointer branches; tarball installs follow GitHub Releases, filtered on the prerelease flag. Full details, including the maintainer release procedure, are in [`docs/release-channels.md`](docs/release-channels.md).
 
 ## What Quern Does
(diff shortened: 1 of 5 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #109, opened 2026-09-05
Track the project conventions instead of keeping them in an ignored file

jerimiah797:
## Why

`CLAUDE.md` was gitignored under a `# Local AI` heading, so everything in it —
architecture, project layout, code conventions, design decisions — existed only
on whichever machine happened to have written it. The other machine working on
this repo could not see any of it, and neither could anyone else: this is a
public repo with no `CONTRIBUTING` file at all.

Nothing in it was private. It was orientation and conventions. The only
AI-specific lines documented `scripts/merge-pr.sh` (already tracked) and the
commit-trailer policy (already enforced by the `commit-msg` hook).

## What changed

- **`CONTRIBUTING.md`** (new) — the durable content, and now the single source of
  truth for it.
- **`CLAUDE.md`** — no longer ignored, cut from 109 lines to 26. It imports
  `CONTRIBUTING.md` and lists only the handful of things that actually cost time
  when got wrong, each *pointing at* a section rather than restating it. Two
  copies of a convention drift, and the copy an agent happens to load wins.
- **`.gitignore`** — `CLAUDE.md` removed; `.claude/` still ignored, since that
  genuinely is machine state.

## What was deliberately dropped rather than moved

**The "Key API Surface" section.** `docs/api-reference.md` already covers it, and
`tests/test_readme_sync.py` checks that file against the routes and MCP tools
actually registered. A hand-maintained third summary would drift silently with
nothing to catch it, so `CONTRIBUTING.md` points at the guarded copy instead.
Same reasoning for the CLI command list, which `README.md` owns and
`test_readme_cli_block_matches_dispatched_commands` guards.

## New content

One section is genuinely new — **error-path conventions**, under *Code
conventions*. Every one of the nine review findings across #106 and #107 landed
in failure handling; not one touched feature logic. Each bullet is anchored to a
specific finding that cost a review round, which is what makes it usable as a
checklist rather than as advice:

- catch base classes, not the subclasses you have seen so far
- a failed check must never read as a passing one
- every failure reaches the exit code, and names the step
- a success marker must not survive a failure, including the path that raises
- tests inject every external lookup; re-audit when a call site moves
- mutation-test the guard — two tests here passed against the bug they claimed
  to cover

## Verification

- **1805 passed**, ruff clean.
- Checked that every convention from the old file survives the move, and that
  `CONTRIBUTING.md` contains zero `api/v1` references, so no third copy of the
  API surface was created.

## Summary by CodeRabbit

* **Documentation**
  * Added a contributor guide covering project architecture, setup, commands, coding conventions, testing, workflows, runtime behavior, and API references.
  * Added guidance for contributors and automated agents on documentation standards, error handling, pull request workflows, commit requirements, and review-script behavior.
  * Documented development environment setup, daemon operation, proxy behavior, design decisions, and common project pitfalls.
  * Updated repository tracking so the guidance file is included in version control.


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #110, opened 2026-09-05
Derive the repository from the git remote instead of hardcoding an owner

jerimiah797:
## The problem

`scripts/pr-review-status.py` and `scripts/merge-pr.sh` both hardcoded
`jerimiah797/quern`. The repo now lives at `quern-dev/quern`, and those calls
kept working **only because GitHub serves a permanent 301 from the old owner
path**:

```
jerimiah797/quern -> HTTP 301, redirects to: repositories/1155621773
full_name: quern-dev/quern
```

So the scripts were reaching the right repository by redirect, not by knowing
where it was.

That redirect is not a guarantee. A repository later created at
`jerimiah797/quern` — a fork, or anyone claiming the name — supersedes it. After
that, `pr-review-status.py` reports on somebody else's PRs and `merge-pr.sh`
attempts to merge into their repository. **Neither would look wrong while doing
it**, which is what makes it worth fixing before it matters rather than after.

## The fix

Both derive `owner/name` from `git remote get-url origin`, and **share one
detection**: `merge-pr.sh` asks the status script via a new `--repo-slug` flag
rather than parsing the remote itself.

That sharing is the load-bearing part. If the gate checked one repository while
the merge ran against another, the check would be meaningless in the most
dangerous way available — it would pass, and the merge would land somewhere
else. One lookup makes disagreeing impossible.

`parse_remote` returns `None` for anything it cannot read rather than falling
back to a default. A guess here sends a merge somewhere nobody asked for.

It also resolves a **fork to the fork**, which the hardcoded value never could —
a contributor no longer has to edit the script to use it.

## Verification

- **1857 passed**, ruff clean.
- End-to-end: the gate still reports correctly against a live PR (#109), and
  `merge-pr.sh` resolves the same slug the gate uses.
- Covers scp-style, https, ssh://, git:// and self-hosted-forge remotes, plus
  six unparseable inputs that must return `None`.
- Mutation-tested: making `parse_remote` guess a default, re-hardcoding the
  owner in `merge-pr.sh`, and running detection at import time are each caught.

## One subtlety

`REPO` is detected only under `__main__`. The tests import this module to
exercise `parse_remote`, and importing it must not shell out to git or
`sys.exit` on a machine with no `origin` remote. There's a test asserting that.

## Summary by CodeRabbit

- **Chores**
  - Repository targeting now adapts to the configured remote instead of relying on a fixed repository.
  - Supports GitHub and self-hosted forge URLs, including SSH, Git, fork, and credential-bearing formats.
  - Reports an error when the repository cannot be determined, helping prevent unintended operations.
  - Added an option to display the detected repository.
  - Sensitive credentials are omitted from repository detection errors.

- **Tests**
  - Expanded coverage for remote parsing, repository detection, host-qualified operations, and merge targeting.

### DIFF
--- a/scripts/merge-pr.sh
+++ b/scripts/merge-pr.sh
@@ -28,6 +28,15 @@ for a in "$@"; do
   esac
 done
 
+# Ask the status script which repository this clone belongs to, rather than
+# naming one here. Both must agree: if the gate checked one repo and the merge
+# ran against another, the check would be meaningless in the most dangerous
+# possible way. Sharing one detection makes disagreeing impossible.
+REPO=$(python3 scripts/pr-review-status.py --repo-slug) || {
+  echo "Could not determine the repository from the git remote."
+  exit 1
+}
+
 if [ -n "$FORCE" ]; then
   echo "Skipping the review gate deliberately (--force)."
 else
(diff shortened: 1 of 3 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

Issue #102, opened 2026-09-04
Network quiescence as a readiness signal, when the proxy is already capturing


jerimiah797:
## Why

`wait_for_settle` (#100) answers *has drawing stopped*. It cannot answer *has
anything arrived*: a blank page part way through a slow load is perfectly still.
Measured against a request to a non-routable address — Safari showed a white
screen with its progress bar stalled, and settle reported `settled: true` after
**1.6s** with zero change between frames. A longer timeout does not help,
because nothing is moving.

The proxy sees what the screen cannot. The two signals are complementary and
neither subsumes the other:

| | answers | catches | misses |
|---|---|---|---|
| **network quiet** | has data stopped arriving | the slow load settle sleeps through | rendering after the last byte; anything with no network at all |
| **screen settled** | has drawing stopped | layout, animation, a modal sliding in | a still, blank, loading page |

The useful condition is both, in that order: data stops, then drawing stops.

## The blocker

**Flows only enter the store when they finish.** `addon.py` writes on
`response()` — *"called when a complete response has been received"* — and on
`error()`. The `request()` hook runs for every flow but only evaluates mocks and
intercept rules; nothing records the flow as in flight.

So a request hanging for 45 seconds is **invisible**: `query_flows` shows
nothing new, which is indistinguishable from an idle network. Built on today's
store, network-idle would fail on exactly the case that motivates it.

The fix is contained — `request()` registers the flow, `response()`/`error()`
clear it, and the store exposes an in-flight count — but it is a prerequisite,
not a detail.

## Design: only when the proxy is already capturing

The system proxy is **off by default** and agents opt in with
`configure_system_proxy`. So this can never be the baseline readiness signal,
and should not try to be:

- **Never turn the proxy on to answer a wait.** Enabling it is a side effect
  with system-wide reach, and a wait is not the place to take it.
- **Degrade to settle-only when it is off**, and say so in the result, so a
  caller can tell "the network was quiet" from "nobody was listening". Silently
  reporting quiet because nothing was being captured is the same fail-open shape
  as #98 and #99.
- **Compose rather than replace.** `wait_for_settle` stays as it is — always
  available, zero setup. Network quiescence is an additional condition on top.

## What it still would not fix

Worth stating so this is not mistaken for the general answer to slow loads:

- **Certificate pinning and non-HTTP traffic never appear**, so "no flows" is
  not "no network" for every app.
- **Some connections never go quiet** — websockets, SSE, long-poll, video.
[...]

PR #97, opened 2026-09-03
feat: identify a screen by the page it is showing

jerimiah797:
The second half of the design discussion. Follows #94 and #96.

## The gap

A screen built entirely on a web view cannot be named by its elements, because
there are none to name — an `SFSafariViewController` reports **one** element to
the accessibility tree, the Application. In Metatext's knowledge base seven
screens carry `landmarks: []` for exactly this reason, and `identify_screen` has
never been able to answer for any of them.

## The fix

A landmark may now match a page URL:

```yaml
landmarks:
  - { web_url_contains: "/settings" }
```

The URL comes from the Web Inspector's **page listing** — one round trip, no
probes. That is what makes it usable for identification, where the dozen
hit-tests a content sweep needs would not be. It is requested only when some
loaded landmark asks for one, so a knowledge base using none pays nothing.

## Three ways it refuses to over-match

- **Combines under AND.** A page open *behind* a different screen does not
  identify that screen as this one.
- **Fails closed.** An `ASWebAuthenticationSession` publishes no pages at all,
  and the listing can fail outright. Neither is read as a match.
- **`absent: true` still inverts**, so "this screen is *not* the settings page"
  works too.

`Landmark.element` becomes optional, which needed a model validator: a landmark
naming neither an element nor a URL can match nothing, and silently treating it
as satisfied would make its screen match **everything**. The parser had its own
check for that, which the validator made redundant — removed, so the rule lives
in one place. Mutation-verified that the validator is what the test pins.

## Verified live

Metatext's Account Settings, whose entire native summary is `"Metatext screen."`:

```
on the page:     identified_as: account-settings | exact
after dismissal: identified_as: None             | none
```

Suite 1618, ruff clean, guards individually mutation-verified.

## Summary by CodeRabbit

- **New Features**
  - Added URL-based screen identification for web views.
  - Web page matching supports required and absent-match conditions, native UI combinations, and optional hosting-process restrictions.
  - Web page URLs are inspected only when needed on supported iOS simulators.
  - Screen identification fails safely when page details are unavailable, including uncertain simulator ownership.

- **Documentation**
  - Clarified background-page behavior, URL specificity, native landmark combinations, process scoping, and platform limitations.

Issue #74, opened 2026-09-01
Retrying a timed-out sim-bridge command can double-act: the command was already written


jerimiah797:
## Summary

`SimBridgeManager.send()` waits up to 30s for a response, but the command has **already been written to stdin** by then. For a non-idempotent command the action may well have happened, and nothing in the response distinguishes *"never sent"* from *"sent, no reply"*.

`server/device/sim_bridge.py`:

```python
line = json.dumps(cmd, separators=(",", ":")) + "\n"
self._process.stdin.write(line.encode())
await self._process.stdin.drain()          # <-- command is now out
try:
    result = await asyncio.wait_for(self._pending_response, timeout=30.0)
except TimeoutError:
    raise RuntimeError(f"sim-bridge command timed out: {cmd.get('cmd')}")
```

A caller that retries on timeout can therefore double-act.

## Why it matters now

The command set splits cleanly, and the split is invisible to the caller:

| Idempotent | Not |
|---|---|
| `describe-ui`, `screenshot`, `probe-point`, `list` | `tap`, `type`, `swipe`, `button`, `home`, `lock`, `siri`, `volumeup`, `volumedown`, `applepay` |

Retrying a timed-out `describe-ui` is free. Retrying a timed-out `tap` may tap twice — and a double-tap is often not a visible failure, just a wrong result: a list item opened and closed, a toggle returned to its original state, a payment sheet dismissed and reopened. That is worse than an error, because it looks like the automation worked.

The contrast with saturation is the useful framing, and it is why this is worth separating out:

- **`SimBridgeSaturatedError` (#68) is safe to retry** — it is raised *before* anything is written, so the command provably did not happen.
- **A timeout is not safe to retry** — the command was written and may have executed.

Both currently surface as "the call failed". An agent, or a `wait_for_element`-style retry loop, has no way to tell them apart.

## Suggested handling

Roughly in order of value:

1. **Make the distinction visible in the type.** A distinct error for "command was dispatched, no response" versus one for "command was never dispatched". Callers can then choose per command; agents can be told the rule in the tool description.
2. **Mark commands idempotent or not** at the point they are defined, so the decision is data rather than tribal knowledge. It is a short, stable list.
3. **Do not auto-retry non-idempotent commands** anywhere in Quern, and say so where retries are implemented.
[...]

PR #111, opened 2026-09-06
Bump to 0.15.0-beta.1, and guard the two places that declare it

jerimiah797:
Cuts the beta for the work merged since `v0.14.1` — **29 PRs, 117 commits**.

## Why minor, not patch

The bulk of it adds capability rather than fixing behaviour:

- **Web content on iOS simulators** (#90–#97, #101, #103, #104) — reading and
  tapping inside `WKWebView`, `SFSafariViewController` and
  `ASWebAuthenticationSession`; screen-settle detection instead of sleeping;
  clearing and verifying text entry.
- **Accessibility bridge recovery** (#89) — self-heals the wedge XCUITest leaves.
- **External tool inventory, updates and service health** (#105–#108) — versions
  and provenance per install *site*, upgrade offers driven by latest rather than
  by a floor, and health checks for tunneld and the mitmproxy system extension.
- **Merge/review tooling** (#86–#88, #110).

MCP tool count moved 107 → 109.

## The guard

The version is declared twice — `pyproject.toml` and `mcp/package.json` — and
nothing compared them. `get_version()` reads the former and calls it the single
source of truth, which is true for the Python server and not for the project:
the npm package carries its own copy.

A release could therefore ship a server announcing one version on `/health`
while the MCP wrapper announced another. Invisible until two machines disagree
and neither can say why — the same failure shape as the tool-version drift
#67 was opened for.

The guard also pins the version's *shape*, since a release is cut from that one
field and a typo there is caught by nothing else.

## Verification

- **1872 passed**, ruff clean.
- Mutation: reverting `mcp/package.json` to `0.14.1` fails the sync test.
- `get_version()` returns `0.15.0-beta.1` end to end.

## Not blocking, but worth knowing

Five issues remain open: #73 (tunneld remediation, has a live repro preserved on
another machine), #74 (sim-bridge retry ambiguity — mitigated, root fix
outstanding), #84, #78, #102. None are regressions from this work, and a beta is
where the remaining ones get exercised. #67 is closed with a writeup of where the
implementation diverged from its proposal.

## Summary by CodeRabbit

* **Release**
  * Updated the project and package version to **0.15.0-beta.1**, marking the latest beta release.

* **Tests**
  * Expanded version validation to support the complete Semantic Versioning 2.0.0 format, including prerelease identifiers and build metadata.
  * Added coverage for valid and invalid version formats, including leading-zero prerelease numbers and empty prerelease components.

### DIFF
--- a/mcp/package.json
+++ b/mcp/package.json
@@ -1,7 +1,7 @@
 {
   "name": "quern-debug-mcp",
-  "version": "0.14.1",
-  "description": "MCP server for Quern — thin wrapper over HTTP API",
+  "version": "0.15.0-beta.1",
+  "description": "MCP server for Quern \u2014 thin wrapper over HTTP API",
   "type": "module",
   "main": "dist/launcher.cjs",
   "bin": {
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "quern-debug-server"
-version = "0.14.1"
+version = "0.15.0-beta.1"
 description = "Debug log capture and AI context server"
 readme = "README.md"
 license = {text = "Apache-2.0"}
--- a/tests/test_version_sync.py
+++ b/tests/test_version_sync.py
@@ -51,8 +51,37 @@ def test_the_server_reports_the_declared_version():
     assert get_version() == _pyproject_version()
 
 
+# A SemVer numeric identifier: no leading zeroes. `\d+` would accept `01.2.3`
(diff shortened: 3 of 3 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #113, opened 2026-09-06
Fix two cross-references the changelog merge broke

jerimiah797:
Found by reading the **published release page**, not the diff. The section merge
in #112 was structurally correct — one heading each, all 21 entries preserved —
but it broke two references that were fine while those entries lived under
`[Unreleased]`.

## Dangling "(below)"

A note said `identify_by:` was retired *"(below)"*. After the merge that entry
sits in `### Changed`, **above** `### Notes`. It also repeated what the Changed
entry already says about existing knowledge bases loading unchanged, so it is
dropped rather than repointed.

## Two notes with no antecedent

These opened with *"The detection is deliberately narrow"* and *"There is no
reload path short of killing the process"*. Both were continuations of the #66
bridge-recovery entry they used to directly follow. Merged into a larger release
they sit several sections from it, so "the detection" refers to nothing. Each now
names its subject.

## Note

The published release body is edited separately — it was generated from the
CHANGELOG before this fix, and the tag is unaffected.

## Summary by CodeRabbit

* **Documentation**
  * Updated the changelog with clearer notes on bridge-restart detection criteria.
  * Documented that terminating and respawning the bridge is the only available recovery mechanism.

* **Tests**
  * Improved checks that keep documented command-line options synchronized with CLI options, including alternate flag spellings.
  * Added validation for options in command examples, inline code, explicitly allowlisted undocumented options, and stale allowlist entries.


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #116, opened 2026-09-07
test(updater): stop reading the developer's channel, and cover beta (#115)

### LINKED ISSUE #115: test_updater.py fails on the beta channel: mocks hardcode origin/release/stable
## Summary

`tests/test_updater.py` fails for anyone whose update channel is `beta`. Five of
its twelve tests fail; on `stable` all twelve pass.

Noted in passing on #114 as "filed separately" — this is that issue, filed after
the fact.

## Reproduction

Isolated so it does not touch a real config:

```sh
TMPHOME=$(mktemp -d); mkdir -p "$TMPHOME/.quern"
echo '{"update_channel": "beta"}' > "$TMPHOME/.quern/config.json"
HOME="$TMPHOME" .venv/bin/python -m pytest tests/test_updater.py -q
```

```
5 failed, 7 passed          # channel = beta
12 passed                   # channel = stable (unset)
```

Failing:

- `test_update_via_git_on_feature_branch_with_updates_warns_and_skips_pull`
- `test_update_via_git_on_feature_branch_with_no_updates_notes_and_skips`
- `test_update_via_git_on_release_branch_with_updates_still_pulls`
- (plus two more)

## Cause

The tests key their mocked `subprocess.run` table on the literal
`origin/release/stable`:

```python
("git", "rev-list", "HEAD..origin/release/stable", "--count"): _make_run(...)
```

`tests/test_updater.py:78`, `:97`, `:109`, `:166`.

The code under test calls the real `_get_release_branch()`, which reads
`~/.quern/config.json`. On the beta channel it asks for
`origin/release/beta`, misses every entry in the table, and the mock returns
whatever its default is.

So the suite's result depends on the developer's machine configuration — the
tests are not hermetic.

## Why it matters more now

v0.15.0-beta.1 is the first release in a while that gives anyone a reason to be
on the beta channel. Every contributor who switches to test it gets five
failures unrelated to whatever they are working on, and the failure text points
at the updater rather than at their own channel setting.

## Fix

Parametrise the mock table on `_get_release_branch()`'s result, or pin the
channel in a fixture so the tests do not read the developer's config at all. The
latter is probably right: a unit test for update mechanics should not vary with
who runs it.

Two of the twelve already do the right thing —
`test_get_release_branch_defaults_to_stable` and
`test_get_release_branch_respects_beta_channel` monkeypatch the channel
explicitly. The other ten inherit it.

jerimiah797:
Closes #115.

Five of `test_updater.py`'s twelve tests failed for anyone who had run `quern set-channel beta`. The mocked command tables are keyed on the literal `origin/release/stable`, while the code under test calls the real `_get_release_branch()`, which reads `~/.quern/config.json`. On the beta channel every key missed.

## Why it read as a product bug

The dispatcher's default is deliberate, and documented as such:

> Default to a successful no-op so unmocked git invocations don't crash the test — they'd surface as the wrong answer rather than an exception, which is what we want when diagnosing a missed mock.

That is sound for its intended case. Here it converted an environmental mismatch into five failures whose text points at the updater rather than at the machine — which is exactly how this got mistaken for a real defect on `fix/tunneld-restart-and-fallback`, where I deselected the file to get a clean run.

## The fix

An autouse fixture pins the channel, so the suite no longer varies with who runs it. The two tests that are *about* the mapping already monkeypatched it for themselves and still override deliberately.

Per the scoping correction on the issue, this tracks the channel **setting**, not the installed version — `set-channel` writes the config and changes nothing else until the next update, so the tests could fail on a checkout that had not moved, while a checkout of the beta tag with the channel unset passed.

## And the coverage whose absence hid it

Every table named `release/stable`, so the one behaviour the channel setting actually controls had **no test at all**. `_check_via_git` is now exercised on beta, asserting it asks about `origin/release/beta`. That is arguably the more valuable half: the bug was possible because the beta path was untested, not merely because a literal was hardcoded.

## Verification

| config | before | after |
|---|---|---|
| Isolated `HOME`, `update_channel: beta` | 5 failed, 7 passed | **13 passed** |
| Isolated `HOME`, channel unset | 12 passed | **13 passed** |
| This machine's real config (beta) | 5 failed | **13 passed** |
| Full suite | needed `--deselect tests/test_updater.py` | **1902 passed**, no deselection |

Mutation-checked by flipping the fixture to `autouse=False`, which reproduces the original five failures exactly.

## Summary by CodeRabbit

- **Tests**
  - Pinned update-channel behavior in tests to avoid results varying with local configuration.
  - Added coverage for beta-channel update checks, including the expected release branch and commit count.

- **Documentation**
  - Added an Unreleased changelog entry describing the test reliability improvements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **`test_updater.py` no longer depends on the developer's update channel** (#115) — five of its twelve tests failed for anyone who had run `quern set-channel beta`. The mocked command tables are keyed on the literal `origin/release/stable`, while the code under test calls the real `_get_release_branch()`, which reads `~/.quern/config.json`; on the beta channel every key missed, and the dispatcher's deliberate default — a silent successful no-op, so a missed mock surfaces as a wrong answer rather than an exception — turned that into five failures pointing at the updater instead of at the machine. An autouse fixture now pins the channel, so the result no longer varies with who runs the suite. This tracks the channel *setting*, not the installed version: `set-channel` writes the config before any beta code exists, so the tests could break on a checkout that had not moved.
+
+
 
 ## [0.15.0-beta.1] - 2026-09-06
 
--- a/tests/test_updater.py
+++ b/tests/test_updater.py
@@ -13,6 +13,27 @@ from unittest.mock import MagicMock, patch
 import pytest
 
 
+@pytest.fixture(autouse=True)
+def pinned_channel(monkeypatch):
+    """Keep these tests off whatever channel the developer happens to be on.
+
+    `_get_release_branch()` reads `update_channel` from `~/.quern/config.json`,
+    so without this the command tables below -- which are keyed on the literal
+    `origin/release/stable` -- miss every entry once someone has run
+    `quern set-channel beta`, and the result of the suite depends on who runs
+    it (#115). The dispatcher defaults a missed key to a successful no-op, by
+    design, so the symptom is a wrong answer rather than an error, and it
+    points at the updater rather than at the machine.
+
+    Note this tracks the *setting*, not the installed version: the channel is
+    written by `set-channel` before any beta code exists, so the tests can
+    break on a checkout that has not moved.
+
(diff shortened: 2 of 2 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

Issue #117, opened 2026-09-07
Revive the menu-bar daemon manager: code is done, blocked on Developer ID creds


jerimiah797:
## Summary

`feat/menubar-daemon-manager` (`8076cde`, on `origin`) is a complete, working menu-bar app that never landed. It stalled in June for want of Developer ID credentials and time, not because of a problem with the code, and was then deprioritised behind the webview work. Recording its state so reviving it does not start with archaeology.

## What is already on the branch

1,027 lines across 14 files:

- **`macos/QuernMenuBar/`** — an Ollama-style `LSUIElement` status-item app (no Dock icon). Swift sources for the status item and menu, a state reader, a CLI wrapper, a self-relaunching updater, and a SwiftUI settings window with an `SMAppService` login item. `Info.plist` template and a `swiftc` `build.sh` producing a universal binary version-stamped from `pyproject.toml`.
- **`scripts/release-menubar.sh`** — build → sign → notarize → staple → assemble asset tarball → upload, documented as step 5 in `docs/release-channels.md`.
- **`server/lifecycle/updater.py`** — `_fetch_latest_release` prefers an uploaded `quern-*.tar.gz` asset over GitHub's source tarball, falling back when absent.
- **`server/lifecycle/setup.py`** — `launch_menubar_app()` opens the bundled app post-install so it self-registers for launch-at-login; no-op for source installs.

Design was Option A deliberately: no Sparkle, no second update path. The signed app ships inside the release tarball and updates through Quern's existing updater. It coexists with `quern start` and MCP `ensure_server`; Quit leaves the daemon running, Opt-Quit stops it. State is read from unauthenticated `~/.quern/*.json`, so it needs no API key.

The commit reports `swiftc` typecheck and universal build succeeding, and the Python suite green.

## The blocker

**The Developer ID for this project is on the other machine (`scorpius`), not the work machine.** Signing and notarisation cannot happen anywhere else, so the release half of this is `scorpius`-only by necessity. Everything else is portable.

## Drift since the branch point

| | |
|---|---|
| Branch point | `396ce57`, 2026-06-05 |
| `main` ahead by | **191 commits** |
| `server/lifecycle/updater.py` | 7 commits, +142 / −27 |
| `server/lifecycle/setup.py` | 5 commits, +186 / −3 |

**The Swift is unaffected.** Everything under `macos/` is new files `main` has never touched, so there is no conflict risk there — the rebase is confined to the two Python files.

[...]

PR #124, opened 2026-09-09
Report proxy and update state the menu bar can trust

### LINKED ISSUE #122: state.json proxy_status is stuck at "starting" — the success transition is never written
## Summary

`proxy_status` in `~/.quern/state.json` is written as `"starting"` at boot and
**never advanced to `"running"`**. Any consumer reading the state file sees a
proxy that is permanently starting up, while the API reports it correctly.

Observed right now, on a proxy that has been serving traffic for three days:

```
API  (/api/v1/proxy/status) ->  running
state.json  proxy_status    ->  starting
```

## Cause

There are only two writers:

- `server/main.py:831` — `"proxy_status": "starting" if enable_proxy else "disabled"`
- `server/lifecycle/watchdog.py:49` — `update_state(proxy_status="crashed")`

Nothing ever writes `"running"`. The field's own type annotation documents a
value the code never produces:

```python
# server/lifecycle/state.py:46
proxy_status: str  # "running", "stopped", "crashed", "disabled"
```

So the state machine is `starting → (crashed)`, with the success transition
missing.

## Why it matters now

It was invisible while the only consumers were authenticated, because
`proxy_status` the *API* computes status live from the adapter and is right.

The menu-bar app (#117) reads the **unauthenticated state files on purpose** —
that is what lets it work without an API key — so it is the first consumer to
surface the stale value. Its menu currently reads:

```
Proxy: starting
```

on a proxy that is running fine. Any future consumer that avoids the API key the
same way inherits this.

## Fix

Write `"running"` once the adapter is actually serving — the same place the
adapter transitions to running and the watchdog begins watching it. Worth
checking `"stopped"` is written on a clean shutdown too, since it is also in the
annotation and may have the same gap.

### LINKED ISSUE #123: Update notification claims an update when a git install is ahead of its channel (cosmetic)
## Summary

Running v0.15.0-beta.1 on the `stable` channel, quern reports an update is
available — **to v0.14.1**, which is older. The persisted result:

```json
{
  "current_version": "0.15.0-beta.1",
  "latest_version": "0.14.1",
  "update_available": true,
  "message": "Update available (v0.14.1) — run \"quern update\" to install it"
}
```

## Cause

`server/lifecycle/update_check.py` performs **no version comparison**. It sends
`version` and `channel` to quern.dev and trusts the answer verbatim:

```python
update_available = bool(data.get("update_available"))
latest_version = data.get("latest_version")
```

The request was `version=0.15.0-beta.1&channel=stable`; the endpoint answered
`update_available: true, latest_version: 0.14.1`. The endpoint is not accounting
for a client that is *ahead* of its channel, and the client has no sanity check
that would catch it.

## Why it appeared now

Before the 0.15.0-beta.1 cut, `main` and the latest stable release were both
0.14.1, so this could not occur. Now anyone developing on `main` while on the
default `stable` channel is running a version newer than their channel's latest,
which is exactly the unhandled case.

**This is the normal state for every contributor**: `main` is 0.15.0-beta.1, and
the channel defaults to stable unless `quern set-channel beta` was run.

## Impact

Mild on the CLI, where it is a message. Worse in the menu-bar app (#117), which
renders it as an action:

```
Restart to Update — v0.14.1
```

A one-click downgrade, presented as an upgrade.

## Suggested fix

Two layers, and the client-side one is worth having regardless of what quern.dev
does:

1. **quern.dev** should not report `update_available` when the client's version
   is newer than the channel's latest. "Ahead of your channel" is a real state
   and deserves its own answer.
2. **The client should not trust it blindly.** A version comparison before
   surfacing the message would have caught this without a server change — and
   the same guard protects against any future endpoint bug or an older
   deployment. The repo already has SemVer parsing in
   `tests/test_version_sync.py` that could move somewhere shared.

Related: #117 (the menu-bar app is where this becomes a clickable downgrade).

jerimiah797:
Closes #122. Closes #123.

Both were found by the menu-bar app on #117, which reads the **unauthenticated
state files** rather than the API — that is what lets it work without an API key,
and it is why it saw things nothing else could. In both cases the API was right
and the file or the notification was wrong, so every authenticated consumer had
been shielded from the defect.

## `proxy_status` never reached "running" (#122)

`main.py` wrote `"starting"` at boot; the watchdog wrote `"crashed"`. Nothing
ever wrote the success value, although `state.py:46` documents it:

```python
proxy_status: str  # "running", "stopped", "crashed", "disabled"
```

So the state machine was `starting → (crashed)`, with the success transition
missing. Observed on a proxy that had been serving traffic for three days:

```
API  /api/v1/proxy/status  ->  running
state.json  proxy_status   ->  starting
```

The addon already emits `started` and `stopped`, and the adapter already had a
handler where both fell through the `else`. Those now write the field.

`stop()` writes it too, deliberately: a hard kill gives the addon no chance to
emit, and the file would otherwise keep claiming the proxy is running.

## The update notification claimed an update when ahead of the channel (#123)

quern.dev answers the sha question with string equality:

```js
update_available: latestSha !== null && clientSha !== "" && clientSha !== latestSha
```

It is not being lazy — `fetchRefs()` gives it branch tips, not a commit graph, so
a Worker cannot tell "ahead" from "behind". A git install on `main` is ahead of
its channel pointer, so it was told an update was available to an **ancestor of
its own HEAD**.

**`_update_via_git` never believed this.** It counts properly and declines:

```
git rev-list HEAD..origin/release/stable --count  ->  0
has_updates = behind_count > 0                    ->  False
return 2   # No update needed
```

So the update was always refused and only the *notification* was wrong. Rather
than add a second mechanism, `check_for_updates()` now consults the reality the
action already computes.

**The guard is one-directional.** A non-git install, or any git failure, returns
`False` and the endpoint is believed. It can only suppress a claim it can
positively disprove — never invent one, never silence a real update.

**Tarball installs are unaffected either way.** `_get_head_sha()` returns `None`
without a `.git` directory, so they never send a sha, and the endpoint is correct
on that path — verified live across all four channel/version combinations.

## Verification

Both confirmed through the app that surfaced them:

```
before:  Proxy: starting        …  Restart to Update — v0.14.1
after:   Proxy: running :9101   …  (item gone)
```

- **1942 passed**, ruff clean.
- Six mutations caught: dropping the `started` write; dropping the `stop()`
  write; letting every status event write `running`; removing the ahead-check;
  letting it fire without a repository (which would break tarball installs);
  and treating a git error as "ahead" (which would suppress real updates).

## Not included

quern.dev still asserts something it cannot know. Fixing it needs a GitHub
compare call per check, and nothing depends on it now that the client is
authoritative — the endpoint's answer is only ever used as a hint the client can
overrule. Left for later rather than folded in here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Update notifications are now suppressed when a Git-based installation already contains the configured channel version, reducing outdated prompts.
  - Proxy lifecycle status remains accurate after startup, shutdown, and forced termination.
  - Unrelated proxy events no longer overwrite the displayed status.
  - Proxy status updates no longer block other asynchronous activity, improving responsiveness during startup and shutdown.

### DIFF
--- a/server/lifecycle/update_check.py
+++ b/server/lifecycle/update_check.py
@@ -85,38 +85,51 @@ def _get_head_sha() -> str | None:
 
 
 
-def _is_ahead_of_channel() -> bool:
-    """Whether the local HEAD is at or ahead of its channel's pointer branch.
+def _is_ahead_of(latest_sha: str | None) -> bool:
+    """Whether the local HEAD already contains ``latest_sha``.
 
-    Only meaningful for git installs; a tarball install has no repository and
-    never sends a sha, so it never reaches here. Counts commits the pointer has
-    that HEAD does not -- zero means nothing to pull, which is what
-    ``_update_via_git`` requires before it will do anything.
+    Answers the question quern.dev structurally cannot: it holds branch refs
+    with no commit graph, so `clientSha !== latestSha` is the best it can do and
+    "ahead" is indistinguishable from "behind" (#123).
 
-    Returns False on any failure, so an unexpected git state falls back to the
-    endpoint's answer rather than silently suppressing a real update.
+    Deliberately does **not** fetch. An earlier version did, and it was wrong
+    twice over: a failed fetch left a stale ref, `rev-list` then reported zero
+    commits behind, and a genuinely outdated install was told nothing — the exact
+    inversion of this function's contract. It also added up to 25s of blocking
+    git to `check_for_updates()`, which `server/main.py` calls synchronously on
+    the foreground startup path.
+
+    Using the sha the endpoint just asserted avoids both. If the object is
+    present locally, ancestry is decidable offline in milliseconds. If it is not
+    present — a shallow clone, a genuinely newer upstream commit, a tarball
+    install — the answer is unknowable here, and False means "believe the
+    endpoint".
+
+    Returns False on anything unexpected, so this can only ever suppress a claim
+    it positively disproves. It can never invent an update or hide a real one.
     """
+    if not latest_sha:
+        return False
     project_root = _find_project_root()
     if project_root is None or not (project_root / ".git").exists():
         return False
     try:
-        from server.lifecycle.updater import _get_release_branch
-
-        ref = f"origin/{_get_release_branch()}"
-        # Fetch first: without it the pointer ref may be stale enough to report
-        # zero commits behind on an install that genuinely has an update.
-        subprocess.run(
-            ["git", "fetch", "--quiet", "origin"],
-            cwd=str(project_root), capture_output=True, timeout=15,
+        # Confirm we actually have the object before asking about ancestry:
+        # `merge-base` on an unknown sha errors, and we must not read that as
+        # a negative answer to a question we never got to ask.
+        have = subprocess.run(
+            ["git", "cat-file", "-e", f"{latest_sha}^{{commit}}"],
+            cwd=str(project_root), capture_output=True, timeout=10,
         )
-        result = subprocess.run(
-            ["git", "rev-list", f"HEAD..{ref}", "--count"],
-            cwd=str(project_root), capture_output=True, text=True, timeout=10,
-        )
-        if result.returncode != 0:
+        if have.returncode != 0:
             return False
-        return int(result.stdout.strip()) == 0
-    except (OSError, subprocess.SubprocessError, ValueError):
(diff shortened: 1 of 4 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

Issue #127, opened 2026-09-10
Bring Android previews into the same window system as iOS


jerimiah797:
## Where we are

iOS previews are windows owned by `Quern Preview.app`, discovered through AVFoundation, tiled by the app, and now opened and closed automatically as devices come and go.

Android previews are a `scrcpy` subprocess (`server/device/scrcpy_preview.py`) that opens its own SDL window titled `Quern: <name>`. It works. It just isn't part of the same surface: it doesn't appear in the Devices menu, doesn't tile with the iOS windows, and doesn't participate in attach/detach handling.

## What is not possible

Rendering scrcpy's video inside our own window. scrcpy draws into an SDL window it owns, and macOS does not allow reparenting another process's window into your view hierarchy. Its `--no-window` flag only pairs with `--record` or `--v4l2-sink`, and v4l2 is Linux-only.

## Option A — manage scrcpy's window (preferred)

Keep scrcpy as the engine, let the preview app place it. scrcpy 4.0 accepts `--window-x`, `--window-y`, `--window-width`, `--window-height`, `--window-borderless` and `--window-title`.

The app would list Android devices in its Devices menu alongside iOS ones, launch scrcpy at the next free tile position, and track the process so a closed window frees its slot. No decoding, no protocol work, and scrcpy keeps doing what it is good at: unlimited duration, low latency, rotation, input injection, audio.

## Option B — decode natively

Verified working on a Pixel 3 XL, with no scrcpy server involved:

```
adb exec-out screenrecord --output-format=h264 -
```

`ffprobe` reports a valid elementary stream, `h264 1440x2960 @ 25fps`. Decoding via VideoToolbox into an `AVSampleBufferDisplayLayer` would put Android in genuinely the same window class as iOS.

The blocker is in `screenrecord`'s own help text:

```
--time-limit TIME
    Set the maximum recording time, in seconds.  Default / maximum is 180.
```

Three minutes per invocation, so a persistent mirror needs restart-stitching with a visible seam. Also higher latency than scrcpy, no input injection, no audio.

## How other tools solve this

Worth recording, because both avoid the embedding problem rather than solving it.

[...]

PR #128, opened 2026-09-10
Screen mirror: build it at setup, ignore the camera, fit the window, follow the devices

jerimiah797:
Four fixes to the screen-mirror path, all found by using it on real hardware: an iPhone 15 Pro and an iPhone 11 over USB, plus a Pixel 3 XL and a Pixel 7 emulator to check what should be ignored.

## The feature was hidden from the people it is for

The menu-bar app offers "Screen Mirror…" only when `Quern Preview.app` exists on disk, and nothing created that bundle until something asked for a preview through the API or an MCP tool. On a fresh install the item was simply absent, and stayed absent until the user had driven a preview from a terminal — which is the thing the menu bar exists to avoid.

It is now built during setup, which costs about 1.5 seconds. Treated as optional the way scrcpy already is: a machine without Xcode Command Line Tools gets a skipped check carrying `xcode-select --install`, and an offer to open the installer. That command hands off to a macOS dialog and returns immediately, so setup says to re-run rather than pretending to wait.

The compile moved into one synchronous function that setup calls directly and the server reaches through a worker thread, rather than growing a second copy of the `swiftc` invocation. That also takes it off the event loop, which it was not before.

## A window showing a black rectangle

A connected iPhone publishes its rear camera as a second external device, so the app opened a window for it.

The obvious filter does not work: a `.continuityCamera` discovery session returns nothing, because the camera arrives as a plain external device. Measured with both phones attached:

```
external/muxed:  J iPhone 15 Pro         modelID=iOS Device   muxed
                 iPhone 11               modelID=iOS Device   muxed
external/video:  J iPhone 15 Pro Camera  modelID=iPhone16,1   video
```

A screen-capture device is muxed and reports the generic model; a Continuity Camera is video-only and reports the hardware model. The video-only discovery branch is kept rather than dropped, since the model ID is the thing actually being asserted.

## Black bars down each side

The window opened at a fixed 400x710, which is 0.563 wide-to-tall against a modern iPhone's 0.462, and `videoGravity = .resizeAspect` letterboxed the difference.

The real size cannot be read at construction time — a screen-capture device advertises one format of 0x0 until it is actually streaming — so the window polls the video port after the session starts and matches the stream. Both test phones settle at 328 wide against a 710 content height. The window's aspect ratio is pinned too, so dragging a corner cannot bring the bars back.

## Unplugged phones left dead windows

A window survived its device, showing a frozen last frame and holding a session bound to something that no longer existed. That also meant replugging could not attach, because the name was still taken.

AVFoundation does post both notifications for these devices, verified by observing an unplug and replug:

```
[22:32:44] DISCONNECTED name=iPhone 11 model=iOS Device muxed=true
[22:32:52] CONNECTED    name=iPhone 11 model=iOS Device muxed=true
```

So the window closes on detach and opens on attach. Attach honours the launch filter: no arguments means "everything", but `ios-preview "iPhone 11"` asked for one device and must not sprout windows for the rest. Interactive mode never auto-opens, because the server decides what is on screen there — it emits `connected` and `disconnected` events instead, and the server drops a disconnected device so its slot and name are freed. `disconnected` is deliberately distinct from `removed`: nobody asked for it.

Every attach and detach also logs one timestamped line before any policy runs. That trace earned its place immediately — four events across two physical replugs read exactly like one replug re-enumerating twice, and I built a debounce for a flicker that did not exist before the timestamps showed two separate cycles. The debounce is not in this PR.

## Verification

- Both iPhones: one window each, no camera window, no bars, 328x742
- Unplug closes the window, replug opens it, confirmed against the app's own log
- A physical Pixel 3 XL on USB is correctly ignored, since Android preview goes through scrcpy
- Setup rebuilds a deleted bundle from scratch, Info.plist and Resources included
- Full suite 1981, up 5. The skipped-versus-failed distinction is mutation-tested

## Not in scope

Bringing Android previews into the same window system is #127, along with what was measured tonight about `adb exec-out screenrecord` and its 180-second ceiling.

## Summary by CodeRabbit

- **New Features**
  - Added automatic screen-mirror app setup during installation, with optional Xcode Command Line Tools guidance.
  - Added real-time detection of connected and disconnected iOS screen-capture devices.
  - Preview windows now open, close, and resize to match device video proportions.
  - Added reliable tracking for preview commands and completion events.

- **Bug Fixes**
  - Improved device detection to exclude unsupported camera devices.
  - Added actionable timeout errors for stalled preview-app compilation.
  - Improved disconnect handling so unavailable devices are not retained as active previews.
  - Repaired incomplete preview app bundles when rebuilding.

### DIFF
--- a/server/device/preview.py
+++ b/server/device/preview.py
@@ -182,9 +182,19 @@ class PreviewManager:
         self._available: list[PreviewDeviceInfo] = []
         self._ready = asyncio.Event()
         self._reader_task: asyncio.Task | None = None
-        # name -> (operation, future). The operation matters: a disconnect
-        # while an add is in flight must fail that add, not complete it.
-        self._pending: dict[str, tuple[str, asyncio.Future]] = {}
+        # name -> (command id, operation, future).
+        #
+        # The id is what correlates a reply with the request that caused it.
+        # The name cannot: a write that times out was still delivered and may
+        # still run, so its late reply would otherwise be matched against
+        # whatever request holds that name next -- a delayed `removed` landing
+        # on a subsequent `add` and reporting a preview the process had just
+        # torn down.
+        #
+        # The operation matters too: a disconnect while an add is in flight
+        # must fail that add, not complete it.
+        self._pending: dict[str, tuple[str, str, asyncio.Future]] = {}
+        self._command_seq = 0
         self._positions: set[int] = set()
         self._stagger_lock = asyncio.Lock()
         self._bundle_path = QUERN_BIN_DIR / APP_BUNDLE_NAME
@@ -256,7 +266,7 @@ class PreviewManager:
             self._reader_task.cancel()
         self._reader_task = None
         # Reject all pending futures
-        for _name, (_op, fut) in self._pending.items():
+        for _name, (_cid, _op, fut) in self._pending.items():
             if not fut.done():
                 fut.set_exception(RuntimeError("Preview process exited"))
         self._pending.clear()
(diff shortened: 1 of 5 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #129, opened 2026-09-10
Split the menu-bar release script into two phases

jerimiah797:
Release tooling, used for real before being committed: `v0.15.0` was built, signed and notarized through this script, and Apple accepted it on the second poll.

## Why split it

The script did everything in one run — build, sign, notarize, staple, assemble the tarball, upload. That puts the slow, externally-dependent step in the middle, so a bad day at Apple's notary service aborts a release *after* the tag has been cut and the GitHub Release created, with nothing to clean up except by hand.

The two halves have different prerequisites, which is what makes the split natural. Building, signing and notarizing need only a version string. Archiving and uploading need the tag and the Release to exist.

```
scripts/release-menubar.sh --app-only 0.15.0   # -> dist/Quern.app
... cut the tag, create the Release ...
scripts/release-menubar.sh --publish v0.15.0
```

The one-shot `release-menubar.sh <tag>` form is unchanged for existing callers.

## --publish verifies rather than trusts

A staged app can come from anywhere, including a previous release. Before uploading, `--publish` checks that the stamped version matches the tag, the signature is valid, a notarization ticket is stapled, and Gatekeeper accepts it — then that the tag exists at all, since the tarball is archived from the tag rather than from `HEAD`.

Version is checked first deliberately: staging one version and tagging another is the mistake this shape invites, and it is the cheapest thing to detect. Verified both guards by pointing `--publish` at a signed-but-unnotarized bundle; it reported the version mismatch against a wrong tag, and the missing ticket against the right one.

## Notarization no longer uses --wait

`notarytool --wait` long-polls a single request that intermittently dies with `NSURLErrorDomain -1001` on this machine. Under `set -e` that aborts the release even though the submission succeeded server-side. Submitting and then polling `notarytool info` makes a flaky poll a retry instead of a failed release.

Same approach and the same reason as `Apps/burrows/scripts/release-macos.sh` in the burrows repo, which documents having hit exactly this.

## Also

The one-shot path stages into `dist/` before uploading, so a failed upload leaves the expensive artifact on disk and `--publish` can finish without going back to Apple.

## Not in this PR

The version bump and CHANGELOG for 0.15.0, which `docs/release-channels.md` says to commit on `main` directly as the release commit.

## Summary by CodeRabbit

* **Chores**
  * Split app release and publishing into separate stages for safer distribution workflows.
  * Improved release reliability with resilient notarization handling, including recovery from timeout responses.
  * Added stronger pre-publish validation for app version, signing, notarization, Gatekeeper acceptance, and team identity.
  * Release versions can be provided with or without a leading “v”.
  * Publishing now requires the specified release tag to exist before packaging and upload.
  * Developer signing configuration is required for all release operations.

### DIFF
--- a/scripts/release-menubar.sh
+++ b/scripts/release-menubar.sh
@@ -2,34 +2,84 @@
 # Build, sign, notarize, and publish the Quern menu-bar app as a release asset.
 #
 # Run on a Mac with the Developer ID Application identity in the keychain.
-# Slots into docs/release-channels.md at the `gh release create` step: after
-# the GitHub Release for the tag exists, run this to attach a tarball asset
-# that bundles a signed+notarized Quern.app alongside the full source tree.
-#
-# The updater (server/lifecycle/updater.py) prefers this asset over GitHub's
-# auto-generated source tarball, so end users get the menu-bar app on update.
+# Slots into docs/release-channels.md at the `gh release create` step: the
+# release asset is a tarball bundling a signed+notarized Quern.app alongside
+# the full source tree. The updater (server/lifecycle/updater.py) prefers it
+# over GitHub's auto-generated source tarball, so users get the app on update.
 #
 # Usage:
-#   scripts/release-menubar.sh <tag>      # e.g. v0.14.0
+#   scripts/release-menubar.sh <tag>                  # everything, one shot
+#   scripts/release-menubar.sh --app-only <version>   # phase 1: app -> dist/
+#   scripts/release-menubar.sh --publish <tag>        # phase 2: archive+upload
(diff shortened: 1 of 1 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #132, opened 2026-09-11
Check certificate trust before enabling capture

jerimiah797:
Removes a gotcha that costs a debugging session every time it happens: ask for the proxy to be enabled, get it, and then watch every network call fail. The device does not trust the mitmproxy CA, so HTTPS from it fails with nothing pointing at the proxy — what you see is a blank screen or an app with no network, and the natural next move is to debug the app.

Both halves of that state were already Quern's. It configures the proxy and it installs the CA; only the correlation was missing.

## What happens now

`configure_system_proxy` preflights booted simulators against `cert-state.json` and refuses with **428** when any of them would fail, naming the devices.

Scoped to simulators deliberately: the macOS system proxy is what they route through, because they share the host's network stack. A physical device is proxied by its own Wi-Fi config and is unaffected by this call, so naming it here would be noise.

The refusal offers three resolutions, not one. A response that only said "install the certificate" would railroad everyone into trusting a MITM root CA, which is precisely the outcome this exists to make deliberate:

```json
{
  "error": "capture_without_cert",
  "devices": [{"udid": "…", "name": "iPhone 16 Pro"}],
  "resolutions": ["install_proxy_cert", "set_auto_install_cert", "skip_cert_check"]
}
```

`skip_cert_check` is the right answer when you are deliberately exercising TLS-failure paths.

## The policy, and why it is visible

`auto_install_cert` answers the question once. Off by default, because installing a root certificate authority is a larger and longer-lived commitment than the proxy toggle that prompts it: it persists across sessions, outlives the capture window, and you have to know it happened to undo it. Anything other than a real boolean reads as unset — a typo in `config.json` should mean "ask me", never consent.

It is surfaced in `proxy_status` **and** in the menu-bar app's Settings pane, both on purpose. A silent, persistent CA-install policy would be worse than the failure it prevents, so the setting is visible where someone would look for it and reversible from the same place. The pane says what the setting costs, not just what it does.

New CLI: `quern set-auto-install-cert [on|off]`.

`proxy_status` also gains a `capture_without_cert` warning, which covers what the preflight cannot reach — an app launched by tapping its icon, or capture enabled before the device booted.

## Agent-facing

The tool description tells agents a 428 here is not transient and must not be retried blindly, and to obtain consent before installing — the same pattern `update_quern` already uses for a consequential action.

## Verified against a running server

| path | result |
|---|---|
| cert present | no refusal, no warning |
| cert missing | 428, device named, three resolutions |
| refusal | configures nothing, `system_proxy_configured` stays false |
| `skip_cert_check` | proceeds, 200 |
| after proceeding | `capture_without_cert` appears in status |

The preflight never fails closed. Blocking capture over a bug in the check would be worse than the failure it prevents, so any error reports nothing missing and lets the call through.

Suite 2006, up 12.

## Not verified

The Settings toggle's **write** path. SwiftUI inside an `NSHostingController` does not expose its controls to AppleScript, which this codebase already documents, so the checkbox cannot be clicked from a script. The read path is verified — changing the setting with the CLI flips the toggle and its caption in the running app — and the command it calls is tested directly, but a human clicking it is the only real proof.

## Background

`docs/proposals/cert-preflight-on-launch.md` is included. It analysed this from a live session on 2026-09-01 and had been sitting on an unpushed local branch. Its status header now records what shipped and what did not: the `launch_app` preflight it proposes is still unimplemented, so §6's known gap is wider than written.

## Summary by CodeRabbit

* **New Features**
  * Added certificate trust checks before enabling capture, identifying simulators that lack the trusted proxy certificate.
  * Added an automatic certificate-install setting in the CLI and menu-bar app Settings; it is off by default.
  * Added certificate status, warnings, and installation-policy details to proxy status responses.
  * Added an option to bypass certificate checks when configuring capture.

* **Bug Fixes**
  * Prevented capture from being enabled when simulator HTTPS traffic would fail due to an untrusted certificate.

* **Documentation**
  * Documented the automatic certificate-install command and certificate preflight behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Quern checks certificate trust before enabling capture** — configuring the system proxy against a simulator that does not trust the mitmproxy CA fails every HTTPS request from it, and the symptom points nowhere near the proxy: a blank screen, or an app with no network. Both halves of that state were already Quern's — it configures the proxy and it installs the CA — and only the correlation was missing. `configure_system_proxy` now refuses in that state and names the devices, rather than creating it and leaving it to be discovered. `proxy_status` reports the same condition as a `capture_without_cert` warning for the cases the check cannot reach.
+- **`quern set-auto-install-cert`** — answers the question once instead of every time. Off by default, because installing a MITM root certificate authority is a larger and longer-lived commitment than the proxy toggle that prompts it: it persists across sessions and outlives the capture window. The setting appears in the menu-bar app's Settings pane and in `proxy_status`, deliberately — a silent, persistent CA-install policy would be worse than the failure it prevents. Pass `skip_cert_check` to configure the proxy anyway, which is the right answer when deliberately exercising TLS-failure paths.
+
 ## [0.15.0] - 2026-09-10
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -348,6 +348,9 @@ quern version                # Print the installed version
 quern update                 # Update to the latest release on your channel and rebuild
 quern update --tools         # Also upgrade external tools quern installed (pipx, brew)
 quern set-channel [name]     # Show or set the update channel (stable / beta)
+quern set-auto-install-cert [on|off]
+                             # Show or set whether the capture certificate is installed
+                             #   automatically, or Quern asks first (default: ask)
 quern uninstall              # Remove Quern and dependencies installed by setup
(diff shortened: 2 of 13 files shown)


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #135, opened 2026-09-11
Install the menu-bar app where it can be found, and actually relaunch it

jerimiah797:
From a real update report: it completed with no failures, and the old menu-bar app kept running. The reporter then quit it and could not find it again to restart.

Two bugs, and the second explains the first.

## An update never replaced a running app

`launch_menubar_app` opened the bundle on every setup, relying on this, which its own comment stated as a reason it was safe:

> LaunchServices activates an existing instance rather than spawning a duplicate

That behaviour is real, and it is exactly the problem. After an update replaces the bundle on disk, `open` activates the *old running binary* and setup reports "Launched" — true, and describing something that did not happen. An update could therefore never deliver a new menu-bar app to anyone already running one.

A running instance is now asked to quit first, with `osascript` rather than a kill so it can tear down its status item cleanly, and the new build is opened after it exits.

## The app was effectively unfindable

It was installed into the release payload directory, which lives under `~/.local`. That is a dot-directory, so Spotlight excludes it. The app did not appear in search or Launchpad, and was in neither place a person looks for an application.

That matters because quitting it is a menu item. Quit it once and there is no supported way back — which is precisely what happened.

It installs to `~/Applications` now. Standard location, indexed by Spotlight, and no admin prompt inside an otherwise unprivileged setup.

`fetch_menubar_app` checks both locations, since after the move the app only exists in the second one and checking the payload directory alone would re-download the asset on every run.

## Test isolation, again

Fixing this surfaced five tests reading the developer's real `~/Applications`. They passed here and would have failed on a machine without the app installed — the same shape as the tunneld plist tests CI caught earlier today. They are isolated now.

## Verification

End to end against the live v0.16.0 asset: fetched, moved to `~/Applications`, still stapled and Gatekeeper-accepted at the new path, and not re-downloaded on a second run. Both guards mutation-tested — opening without quitting first, and leaving the app in the payload directory, each fail their test.

Suite 2040.

## For anyone already stuck

The app is at `~/.local/share/quern/Quern.app`. Until this ships:

```sh
pkill -f QuernMenuBar && open ~/.local/share/quern/Quern.app
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Menu-bar app updates now ask any running instance to quit before replacing and launching the new version.
  - The menu-bar app is now installed in `~/Applications`, providing a more accessible and consistent installation location.
  - Updates avoid unnecessary downloads when the app is already available in its delivery or installation location.
  - Installation and update behavior now handles source-only setups without attempting an invalid launch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **An update never replaced a running menu-bar app** — setup launched it with `open`, which activates a running instance rather than starting the new binary. So after `quern update` the old build kept running while setup reported "Launched", which was true and described something that had not happened. A running instance is now asked to quit first.
+- **The menu-bar app is installed where you can find it** — it lived in the install payload directory under `~/.local`, a dot-directory Spotlight excludes. It did not appear in search or Launchpad, and was in neither place a person looks. Since quitting it is a menu item, that made it effectively unrestartable. It now installs to `~/Applications`.
+
 ## [0.16.0] - 2026-09-11
 
 ### Added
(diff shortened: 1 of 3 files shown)


jerimiah797:
@coderabbitai review

PR #136, opened 2026-09-11
Fix the menu bar's CLI resolution, update check and silent failures

jerimiah797:
Reported from a git-clone install where the menu-bar app was installed before `quern setup` had written `~/.local/bin/quern`: Start did nothing, "Restart to Update" did nothing, the icon stayed dimmed. Three separate bugs, all sharing one habit — a failed operation that looks exactly like an idle one.

### `installDir` has been wrong since v0.16.0

`QuernCLI.installDir` was "the parent of `Quern.app`". True while the bundle lived inside the install tree; wrong the moment v0.16.0 moved the app to `~/Applications` so Spotlight and Launchpad would find it. The rule kept resolving, to `~/Applications`, so everything derived from it pointed at a directory that will never hold what it wanted. Replaced with an explicit `~/.local/share/quern`.

### "Restart to Update" could never succeed

`Updater.installedVersion()` read `pyproject.toml` from that directory, so it has returned `nil` for every user since v0.16.0. The poll compares against a baseline, and `nil` never differs from `nil` — so `quern update` ran, worked, and the app sat through its full 180-second timeout before reporting one.

Now it asks `quern --version`. That is a subprocess, so the call is async; each tick skips rather than queues while one is in flight, because the venv is being rebuilt mid-update and a call can hang. `nil` means "could not tell, keep waiting", never "unchanged".

### Lifecycle actions swallowed the exit code

Start, Stop and Restart discarded the status and only refreshed. A failure redrew an identical menu, so the click read as a dead menu item. They now raise the same alert `quitAndStop` already used.

### `resolve()` fell back to a guaranteed 127

It ended with `/usr/bin/env quern`. A GUI app does not inherit your shell's PATH, so the clone directory this user had on theirs was invisible to it, and 127 is indistinguishable from the command existing and failing. It now returns `nil` and the app names the wrapper it looked for.

### Verification

A probe compiled against these sources, run against the real filesystem:

| wrapper state | `resolve()` | `installedVersion` |
| --- | --- | --- |
| present and working | path to wrapper | `0.16.1` |
| present, exits nonzero | path to wrapper | `nil` |
| present, prints something else | path to wrapper | `nil` |
| absent | `nil` | `nil` |

Full Python suite passes (2048), ruff clean, universal Swift build clean.

Note this only reaches users through a signed release, not through `quern update` of the server.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Restart to Update by checking the installed CLI version reliably.
  * Start, Stop, and Restart now display an alert when an operation fails.
  * Missing CLI installations now show a clear setup message instead of failing silently.
  * Improved detection of installed releases and command-line tools.
  * Settings now loads and displays the installed version more reliably.
  * Added clearer update progress and failure reporting.


jerimiah797:
@coderabbitai review

Issue #137, opened 2026-09-11
The menu-bar app has no test target, so none of its guards can be mutation-tested


jerimiah797:
`CONTRIBUTING.md` requires that a guard be mutation-tested: revert the fix and confirm a test actually fails. `macos/QuernMenuBar/` has `Sources/`, `Assets/` and `build.sh`, and no test target. So that rule currently applies to half the codebase.

Two agent reviews of #136 both raised it, and #136 is the reason it now matters. That branch moved a lot of state into the app:

- the updater's poll loop, with a wall-clock deadline, an in-flight guard and a re-entrancy guard
- `confirmStartFailed`, a bounded retry chain driven by `DispatchQueue.main.asyncAfter`
- `lifecycleBusy` / `lifecycleFailed`, which gate menu contents and the icon
- `pidPlausiblyOurs`, comparing a process start time from `sysctl` against a recorded timestamp
- `QuernCLI.resolve()` and its watchdog deadlines

The Python half of that branch has tests, and every guard was reverted individually to confirm one fails. The Swift half was verified by running the real app against stub CLIs and reading screenshots. That caught real bugs — including a tint that silently did nothing — but it is manual, slow, and proves nothing on the next change.

Worth noting what the manual checks did catch, because it sets the bar: `contentTintColor` had no effect on a menu bar template image, which no amount of reading the code would have shown. Some of this genuinely needs to render.

## The constraint

`build.sh` invokes `swiftc` directly. There is no SwiftPM manifest and no Xcode project, deliberately — the app builds with the toolchain already present for the rest of the project. A test target usually means adopting SwiftPM or XCTest, which is a real change to how this thing is built and shipped, and the signing and notarisation path in `scripts/release-menubar.sh` depends on the current layout.

So this is not "add a test target", it is "decide what testing this app is worth". Options, roughly in order of cost:

1. **Split the logic that has no AppKit dependency into a plain module and test that.** Version parsing, `pidPlausiblyOurs`, `resolve()`, the state machine transitions. Would cover most of the list above.
2. **Adopt SwiftPM with a test target**, keeping `build.sh` as the release path.
3. **Leave it, and keep verifying by hand**, with the trade-off written down rather than implied.

Needed before any of that: whether the CI runner can run tests that need a display or a run loop, and what this does to the notarised bundle.

jerimiah797:
Worked through this after #136 landed. Two facts make it cheaper than the issue body assumed, and one makes it more expensive.

[...]

PR #138, opened 2026-09-11
Give the menu bar a clock it does not own, and tests that use it

jerimiah797:
Closes the first half of #137.

The question that decided the shape of this was not "can we run Swift tests" but "would they have caught what the reviews caught". Every timing defect found in #136 needed three minutes of real time to reproduce, so none of them was ever reachable by a test that waits. The harness was never the work. The seam was.

## The seam

`Scheduler` is two ways to schedule and one way to read the clock, which is the whole of what this app does with time. `SystemScheduler` is `Timer` in `.common` mode and `asyncAfter`, exactly as before. `TestScheduler` moves only when a test moves it, stepping to each due time rather than jumping to the end, so a body that reads `now` sees the time its own tick represents.

`Updater` takes its scheduler, its two CLI calls and its relaunch as dependencies with real defaults, so nothing but a test passes any of them. Closures rather than a protocol: there are four, they have nothing in common, and a protocol would only give them a shared name. `relaunch` is injectable because the alternative is an `NSWorkspace` call that quits the app mid-test.

## The tests

Seven, each one a defect found by reading during #136:

| test | the defect it pins |
| --- | --- |
| a changed version relaunches into it | the happy path, plus no timer left running |
| an unreadable baseline does not start the update | — |
| a nil baseline cannot be mistaken for a version change | relaunching into the bundle being replaced |
| a hung version check still reaches the deadline | the in-flight guard skipping the only deadline check |
| an unchanged version is not reported as a timeout | reporting failure for an update that worked |
| a failed update stops and reports | — |
| a second click while one is running is ignored | two `quern update` runs against one install tree |

Four mutations, each caught:

```
deadline back inside the completion   6 passed, 2 failed
nil baseline accepted                 5 passed, 2 failed
re-entrancy guard removed             6 passed, 1 failed
polling after a failed update         6 passed, 1 failed
restored                              7 passed
```

## Why not SwiftPM

XCTest means SwiftPM, which wants `Sources/<Target>/`. That forces splitting these sources into a library and an executable, and `build.sh` and `release-menubar.sh` both depend on the current layout. Restructuring the two scripts that produce a signed artifact is a poor trade for test discovery.

`run-tests.sh` compiles `Sources/` minus its `main.swift` against `Tests/` with plain `swiftc`, using the same deployment target `build.sh` uses. That last part is not cosmetic: several SwiftUI modifiers here are deprecated after macOS 13, so the host SDK turns `-warnings-as-errors` into a wall of failures about code the shipped build compiles cleanly.

CI runs it as a step on the `macos-latest` job that already builds the app, so it costs no new infrastructure.

## Worth knowing

Two tests failed on the first run and the rig was at fault both times: it discarded the `Updater`, and the poll body captures `[weak self]` and cancels when that is nil. The test now retains one the way `AppDelegate` does. That is a real requirement the tests surfaced, not bookkeeping.

Rendering is not covered and the README says so. The bug where `contentTintColor` silently did nothing to a menu-bar template image was found by screenshotting the real menu bar and measuring pixels. A headless runner would not have caught it and should not claim to.

Still open on #137: the `AppDelegate` retry chain, which has the same shape and is not yet on the seam.

## Summary by CodeRabbit

- **Bug Fixes**
  - Update polling now continues reliably while menus or modal windows are open.
  - Improved handling of unreadable, unchanged, or delayed update checks.
  - Failed updates stop cleanly, and repeated update requests are ignored while one is already running.

- **Tests**
  - Added automated coverage for update-checking and relaunch scenarios.
  - CI now runs the menu-bar application’s test suite.

- **Documentation**
  - Added instructions for running the menu-bar tests locally and details about CI coverage.


jerimiah797:
@coderabbitai review

PR #146, opened 2026-09-13
Put the Settings window on one gutter, and stop the UDID overflowing

jerimiah797:
Cosmetic, found by eye and fixed against a running signed build over five rounds of screenshots.

## Four gutters in one window

| Where | Inner padding before |
|---|---|
| Server, Proxy, Active device (all use `grid()`) | none |
| Network capture | 6 |
| Updates | none |
| The toggles and button at the bottom | outside any box, so aligned to box *frames* |

Now: one named constant for the padding inside a box, applied in the three places that have to agree, and a second for the bottom group — a different distance, since it also covers the box's own inset.

## The channel picker

Moved too, and this one was a pre-existing bug rather than a knock-on:

```swift
.frame(maxWidth: 220)   // centres by default
```

The segmented control measures about 115 in a 220 box, so it sat ~50pt right of where its label column ends — clear of the value column every row above aligns on. It only became visible once the box around it gained padding. Now `.frame(maxWidth: 220, alignment: .leading)`.

## The UDID overflow

A real 36-character simulator UDID does not fit the value column. It wrapped to a second line while nothing reserved the height for it, so the tail rendered past the GroupBox and over the section below. Clicking the text appeared to cause it; it only forced a re-layout that made it visible. The overflow was always there, and a synthetic short UDID (`ABCD-1234`, left in a state file during development) had been hiding it.

`.fixedSize(horizontal: false, vertical: true)` on the value, so the row takes the height it needs — the same modifier the certificate explanation in this window already uses. Wrapping rather than truncating, because a UDID you can read and select is worth more than one that fits on a line.

## Worth a reviewer's attention

The bottom inset took four passes, and the first three failed for a reason worth recording. A stray brace had attached the padding to the whole body rather than to that group, so every value shifted the entire window and left the relative gap untouched. Three guesses "had no effect" and I reached for a bigger number each time instead of reading that as evidence about the wiring. With it attached correctly the value is just the measured gap — 11 points, taken from the leftmost blue pixel of each checkbox glyph in a screenshot.

Both magic numbers carry comments recording how they were arrived at, since neither is derivable: a GroupBox's own inset is not a published constant, and reading it out of SwiftUI at runtime would be a worse dependency than a number with an explanation.

No behaviour change, so no new tests. 2232 Python and 93 Swift tests unchanged.

## Summary by CodeRabbit

* **UI Improvements**
  * Improved Settings window spacing and alignment for a more consistent layout.
  * Long setting values now wrap within their containers instead of extending beyond them.
  * Refined the placement of channel selection, toggles, and the Documentation button.

### DIFF
--- a/macos/QuernMenuBar/Sources/SettingsWindow.swift
+++ b/macos/QuernMenuBar/Sources/SettingsWindow.swift
@@ -348,6 +348,40 @@ struct SettingsView: View {
     private var u: UpdateInfo { model.snapshot.update }
     private var d: ActiveDevice { model.snapshot.device }
 
+    /// Padding inside a GroupBox, on top of whatever the box insets itself.
+    ///
+    /// Applied by hand because the boxes have to agree and SwiftUI will not
+    /// make them. Before this there were three different gutters in one
+    /// window: one box padded by 6, another not padded at all, and the three
+    /// built from `grid()` not padded either.
+    private static let contentInset: CGFloat = 6
+
+    /// How far the bottom group sits in, to line its checkboxes up with the
+    /// ones inside the boxes above.
+    ///
+    /// A separate number from `contentInset`, and larger, because it covers a
+    /// different distance: a GroupBox's own inset *plus* the padding we add on
+    /// top of it. The two toggles and the button down there sit outside any
+    /// box, so nothing aligns them for free -- and being the only controls in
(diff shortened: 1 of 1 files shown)


jerimiah797:
@coderabbitai review

Issue #147, opened 2026-09-13
DeviceState is simulator vocabulary, and carries no meaning for physical devices


jerimiah797:
`DeviceState` is `booted / shutdown / booting / unauthorized` — simctl's vocabulary with adb's `unauthorized` bolted on. It fits simulators. It does not fit physical devices, and the backends that report them each work around that differently.

## The evidence

**`usbmux.py:137` does not compute it.** Every device it lists gets `state=DeviceState.BOOTED` and `is_connected=True`, unconditionally. There is nothing to compute, so it declares a constant.

**`devicectl.py` invents it from reachability:**

```python
# tunnelState values: "connected" (active tunnel), "disconnected"
# (reachable but no tunnel yet — e.g. WiFi devices), "unavailable".
if tunnel_state in ("connected", "disconnected"):
    state = DeviceState.BOOTED
```

That is a mapping from "can I see it" to "is it running", which are different questions.

**`cert_preflight.py:47` guards before using it:**

```python
if d.device_type != DeviceType.SIMULATOR or d.state != DeviceState.BOOTED:
```

The simulator check comes first. The state check is only meaningful once it has.

So for a physical device the field is either a constant or a guess, and at least one consumer already knows not to trust it without qualification.

## Why it matters, concretely

A physical iPhone reachable over WiFi reports `tunnelState: disconnected` and carries no `bootState` at all. Quern calls it `booted`. It can probably be deployed to. Whether quern can drive it — WDA, device logs, crash reports — is a different question: logs resolve a tunnel first for iOS 17+, and `tunneld` exists precisely to provide RemoteXPC tunnels for developer services.

Both answers are wrong in a way that costs something:

- Call it `shutdown` and a usable device disappears from the list.
- Call it `booted` and an agent gets confusing failures from something quern said was ready.

There is no right answer because the question does not apply. What actually matters for a physical device is already tracked separately or not at all: `is_connected`, paired, and whether a tunnel exists.

## Not urgent, and worth doing deliberately

Nothing is broken today that this would fix — it is why a bug in this area was awkward to fix rather than a bug itself. It touches an enum used by four backends and several consumers, so it wants doing on its own rather than riding along with something else.

Rough shape, not a proposal: either a separate readiness field that means "quern can drive this", computed per device kind, or accept that `state` is simulator-only and have physical-device consumers use `is_connected` and tunnel presence instead.

## Where this came from

[...]

PR #148, opened 2026-09-13
Stop devicectl reporting simulators as physical devices

jerimiah797:
From an external report against 0.16.1: `get_screen_summary` returns **HTTP 500 against a perfectly healthy booted simulator**.

## What happens

Xcode 26 registers simulators as CoreDevices, so they appear in `devicectl list devices` alongside real hardware. `list_devices` appended every paired, reachable device with a hardcoded `DeviceType.DEVICE` and never read `hardwareProperties.reality`, which is exactly `"simulated"` or `"physical"`.

The damage happens one layer up, in `controller.list_devices`:

```python
for d in sim_devices:
    self._device_type_cache[d.udid] = DeviceType.SIMULATOR
for d in physical_devices:
    self._device_type_cache[d.udid] = DeviceType.DEVICE   # overwrites
```

simctl fills the cache first, devicectl second and unconditionally. Any UDID in both lists ends up typed `device` — which is every simulator, once they started appearing in devicectl. `resolve_device` then writes `{"type": "device"}`, UI reads route to the WDA client instead of sim-bridge, and with no WDA build present that is a hard failure. Passing `type: "simulator"` explicitly does not override it.

## The fix

**devicectl skips simulated devices.** simctl is authoritative about simulators — it knows the runtime and the real boot state — and a second, worse entry for the same UDID can only cause this.

**The controller's cache believes the backend** rather than the loop the device arrived in, so the mistake cannot be reintroduced by anything else returning one.

Only an explicit `"simulated"` filters. Older devicectl versions omit the field, and dropping those would make physical devices vanish — a worse failure than the one being fixed.

## Deliberately not changed: the physical-device state mapping

The report also noted shut-down simulators arriving as `booted`, because they report `tunnelState: disconnected`. That is a symptom of simulators being in this list at all, and the filter is the fix.

Narrowing the rule instead reported a **wifi-connected iPhone as shut down** — measured on this machine: it carries no `bootState` and a disconnected tunnel while plainly being awake and deployable. An existing test correctly defended the current behaviour and caught the attempt.

The deeper problem is that the question does not apply: `booted` is simctl's vocabulary, and `usbmux.py` fills the same field with a constant because there is nothing to compute. Filed separately as #147.

## Honest limitation

**I could not reproduce the primary symptom here**, despite matching the reporter's Xcode 26.6 and macOS 26.6.2. Cross-referencing by UDID, none of the three devicectl entries on this machine match any of the 33 simulators simctl knows about — all three are genuinely physical and correctly labelled `physical`.

So the fix comes from the data and the reporter's evidence, not from a local repro. What I could verify locally: physical devices are still listed and unchanged, and the booted simulator types as `SIMULATOR` end to end through the controller.

## Also included

One commit from the quern.dev update-check work: a note on `_get_local_version` explaining why the version is scraped from the raw `pyproject.toml` line rather than read through `importlib.metadata`. The normalised PEP 440 form is not semver — `0.14.1-beta.2` becomes `0.14.1b2` — which the update endpoint would fail to parse, silently restoring the bug where tarball installs never hear about a release. It reads like something to tidy up, so it is written down at the function that would be tidied.

2237 tests. Three mutations, three caught.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed device listings so simulators are consistently identified as simulators when they appear in multiple device sources.
  - Prevented simulator devices from appearing in physical-device results, including simulators registered as CoreDevices with Xcode 26.
  - Preserved correct handling of physical devices and device connection states.

### DIFF
--- a/server/device/controller.py
+++ b/server/device/controller.py
@@ -380,7 +380,13 @@ class DeviceController(DeviceControllerUI):
         for d in sim_devices:
             self._device_type_cache[d.udid] = DeviceType.SIMULATOR
         for d in physical_devices:
-            self._device_type_cache[d.udid] = DeviceType.DEVICE
+            # The backend's own classification, not a hardcoded DEVICE. This
+            # loop runs after the simctl one, so stamping DEVICE here overwrote
+            # the correct SIMULATOR for any UDID appearing in both -- which is
+            # every simulator, once Xcode 26 started registering them as
+            # CoreDevices. devicectl now filters simulators out, and this stops
+            # the mistake being reintroduced if anything else ever returns one.
+            self._device_type_cache[d.udid] = d.device_type
             if d.os_version:
                 self.wda_client._device_os_versions[d.udid] = d.os_version
             if d.name:
--- a/server/device/devicectl.py
+++ b/server/device/devicectl.py
@@ -125,11 +125,40 @@ class DevicectlBackend:
             raw_transport = connection_props.get("transportType", "")
             connection_type = "usb" if raw_transport == "wired" else raw_transport
             tunnel_state = connection_props.get("tunnelState", "")
-            device_type_str = dev.get("hardwareProperties", {}).get("deviceType", "")
+            hardware_props = dev.get("hardwareProperties", {})
+            device_type_str = hardware_props.get("deviceType", "")
+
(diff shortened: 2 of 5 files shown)


jerimiah797:
@coderabbitai review

Issue #160, opened 2026-09-14
Leaked 'usbmux forward --udid test-udid' process squats port 18100 and silently disables physical-device automation


jerimiah797:
A `pymobiledevice3 usbmux forward` process from a test run outlives the test and keeps holding the WDA forward port. On this machine it had been running for **13 days, 18 hours**:

```
$ ps -o pid=,etime=,command= -p 77008
77008 13-18:42:02 .../pymobiledevice3 usbmux forward 18100 8100 --udid test-udid

$ lsof -nP -iTCP:18100
Python  77008 jham  6u  IPv4  TCP 127.0.0.1:18100 (LISTEN)
```

Note `--udid test-udid` — the literal placeholder, so this came from the test suite, and it forwards to a device that does not exist.

## Impact

Port 18100 is bound, so quern's own forward cannot establish. The WDA health check polls `127.0.0.1:18100`, gets nothing, and reports:

```
[wda] Auto-started WDA driver on B34C4EE9 but it did not become responsive.
Check log: ~/.quern/wda/runner-B34C4EE9.log
```

That message sends you to the runner log, where **WDA is running perfectly**. It was reachable the whole time:

```
$ curl http://192.168.1.160:8100/status
{"value":{"build":{"version":"11.4.0", ...
```

Killing PID 77008 restored automation instantly — no restart of anything else.

The failure is silent, permanent until someone kills the process by hand, survives reboots of the quern server, and points diagnosis at the wrong subsystem. It cost a fair amount of time before `lsof` on the port gave it away.

## Suggestions

- Ensure tests tear down forward subprocesses (context manager / fixture finalizer), and never use a placeholder like `test-udid` for something that opens a real listening socket.
- Before binding, check whether the port is already held and by what. If the holder is a `usbmux forward` for a udid that is not a real device, say so explicitly rather than timing out.
- Fold this into the WDA health-check error message: "port 18100 is held by PID N (`--udid test-udid`)" is instantly actionable, where "WDA did not become responsive" is not.

Related: #159, a different reason WDA startup can fail while the runner itself is healthy.

jerimiah797:
**Live on this machine right now, and in a broader form than filed.** Measured
today while verifying triage, with physical devices attached:

    PID    AGE           FORWARD
    48853  6d 23h 11m    forward 18100 8100 --udid B34C4EE9  (iPhone 11)
    48883  6d 23h 11m    forward 18101 8100 --udid B34C4EE9
[...]

Issue #159, opened 2026-09-14
WDA startup reinstalls the runner every time, so a flaky install channel blocks automation on an already-installed runner


jerimiah797:
Starting the WDA driver goes through `xcodebuild test-without-building` (`server/device/wda.py`, driver lifecycle section around line 635; `server/api/wda.py:108`). That command reinstalls the runner app on every start.

When the device's install channel is unhealthy, the install step fails and WDA cannot start **even though the runner app is already installed and perfectly usable**:

```
Testing failed:
	Failed to install the app on the device.
	WebDriverAgentRunner-Runner encountered an error (Failed to install or launch the test runner.
	(Underlying Error: Failed to install the app on the device.
	(Underlying Error: Connection with the remote side was unexpectedly closed :
	"XPCErrorDescription" => "Connection interrupted"
	. Unhandled error domain IXRemoteErrorDomain, code 6)))
** TEST EXECUTE FAILED **
```

Meanwhile the app is present and the device is otherwise reachable:

```
$ xcrun devicectl device info apps --device <hw-udid> | grep -i quern
QuernDriver   dev.quern.driver.xctrunner   1.0   1

$ xcrun devicectl device info lockState --device <hw-udid>
• passcodeRequired: false
• unlockedSinceBoot: true
```

Quern retried the auto-start repeatedly over ~10 minutes and every attempt timed out at the 30s mark:

```
[INFO] quern-debug-server.wda-client: WDA not reachable on B34C4EE9, auto-starting driver...
[INFO] server.device.wda: Cleaning stale WDA runner for B34C4EE9 (pid 7793)
[INFO] server.device.wda: Starting WDA driver for B34C4EE9 (hw_udid=...)
[WARNING] server.device.wda: WDA did not become responsive within 30s for B34C4EE9
```

Every `press_button` / `tap` / `swipe` returned HTTP 500 for the duration.

## The install is not needed

An `.xctrunner` app cannot simply be launched — `devicectl device process launch dev.quern.driver.xctrunner` reports success but nothing comes up, because XCTest needs `testmanagerd` to attach and drive the bundle.

But `pymobiledevice3` implements that harness, and it does not install anything:

```
pymobiledevice3 developer dvt xcuitest dev.quern.driver.xctrunner --udid <hw-udid>
```

WDA answered `/status` **8 seconds later**, after `xcodebuild` had failed for ten minutes straight:

```
[IDE] testRunnerReadyWithCapabilities: {...}
[...]

PR #161, opened 2026-09-14
docs: how to start WDA without reinstalling the runner

jerimiah797:
Two different failures both surface as "WDA did not become responsive", and `docs/guides/ios-wda.md` only covered one of them.

A build or signing problem wants a rebuild. An **install-channel** failure must not get one — the runner is already installed and healthy, only the install step is failing, and every retry fails identically. Hitting this in practice cost a long detour into rebuilding WDA when the runner was fine the whole time.

Changes:

- **How the Driver Works** now states that step 2 reinstalls the runner on every start. That is the fact which makes the failure mode possible, and it was not written down anywhere.
- **Common Failure Patterns** gains two rows: the `IXRemoteErrorDomain` / "Connection interrupted" install failure, and "responsive check fails but the runner log is clean", which is usually something local holding the WDA port.
- New **Starting WDA without reinstalling** section covering `pymobiledevice3 developer dvt xcuitest dev.quern.driver.xctrunner`, which drives the runner through `testmanagerd` with no install step. Measured: WDA answered `/status` in 8s after `xcodebuild` had failed for ten minutes straight, and Quern resumed on the next command with no server restart.
- Warns against `devicectl device process launch` on the runner, which reports success and does nothing, because an `.xctrunner` app is a stub that needs `testmanagerd` to drive it. Easy to lose time on.

Docs only, no code changes.

Refs #159 (the startup path itself), #160 (the stale forward, which is the other half of that new table row).


jerimiah797:
@coderabbitai review

Issue #165, opened 2026-09-14
Installing an app on a proxied device should offer to capture it


jerimiah797:
Raised while demonstrating the cert-consent work. The observation: **reinstalling
an app is the moment its capture is decided**, and right now those two steps are
unconnected.

## The gap

`local_capture` is host-side config naming process names. It survives a
simulator erase — but the *app* does not. So after `simctl erase`, the list
still names `Metatext` while nothing by that name can ever run again. Capture
looks configured, is accepted, and quietly produces nothing from that app.

Reinstalling is when this gets fixed, and it is also when quern has the
information: `install_app` knows the bundle, and `proxy_status` knows whether
this device is being captured. Neither tells the other.

The reverse gap is the same shape: install an app on a device that *is* being
proxied, and nothing suggests adding it to the list. The user has to know that
`local_capture` exists, that it is per-process, and that the process name is not
the bundle ID.

## Why "just add it automatically" is not obviously right

- `local_capture` names **process names**; `install_app` takes a **bundle
  path**. On this machine Metatext's `CFBundleIdentifier` is
  `org.arctian.metatext` while its `CFBundleExecutable` is `Metatext` — the
  executable name is what the proxy wants, and it is available in
  `simctl listapps`, but the mapping is not the obvious one.
- Capturing an app is not free: it routes that app's TLS through the proxy, and
  a device that does not trust the CA then fails every HTTPS request. That is
  the exact failure the 428 gate exists to prevent, so adding to the list
  silently would reintroduce it from a new direction.
- Not every installed app should be captured. Installing a dependency or a test
  harness should not start intercepting it.

So the likely shape is **offer, not do** — the same posture as
`auto_install_cert`: report that this device is being captured and this app is
not in the list, name the process name to add, and let the caller decide. With
an opt-in setting for people who want it automatic.

## Adjacent, and already fixed

`set_local_capture` replaced the list silently, so a caller passing a subset
deleted the rest with nothing said. Now reported — see
`fix/verification-clobbers-the-record`. That is the "how do I not lose the list"
half; this issue is the "how does the list stay true to the device" half.

## Prior art in the tree

- The 428 refusal in `refusal_detail` is the pattern to follow: name the
  problem, name every resolution, change nothing.
- `auto_install_cert` is the precedent for an opt-in that makes the offer
  automatic, including its default-false rule that anything unclear reads as
  "ask me".

Issue #149, opened 2026-09-13
Certificate trust is one boolean over three device kinds that verify differently


jerimiah797:
Not urgent, and not a bug on its own. It is the shape that produced three separate bugs in one sitting, and two of my own fixes for them had to be walked back. Recording the structure rather than patching around it again.

## What the structure is

`cert_installed` is a single boolean in `cert-state.json`, shared by three device kinds whose truth is established in completely different ways:

| Kind | How trust is verified | What invalidates it |
|---|---|---|
| Simulator | SQLite query against `CoreSimulator/Devices/<udid>/.../TrustStore.sqlite3` | an erase |
| Android | `adb` against the system cert store | a wipe |
| Physical iOS | successful intercepted HTTPS flows from a recorded `client_ip` | an erase **or a network change** |

That last column is the part that keeps biting. A WiFi change is inconsequential for a simulator, which routes through the host's network stack, and crucial for a phone, which carries its own per-network proxy config. The same event means nothing to one kind and invalidates the other.

## Three specific things that follow from it

**1. `is_cert_installed` dispatches on Android and assumes simulator for everything else.**

```python
if controller._is_android(udid):
    return await _is_cert_installed_android(...)
# ...everything else goes to the simulator TrustStore path
```

A physical iOS device falls through to `get_truststore_path(udid)`, which builds a path under `CoreSimulator/Devices/`. It does not exist, so the answer is `false` for a phone that genuinely trusts the CA. The name gives no hint that it only answers for two of the three kinds.

**2. A read-shaped call writes to disk.** `is_cert_installed` ends its slow path with `update_cert_state(udid, ...)`. So calling what looks like a query mutates the shared record — and in the physical-device case above, overwrites a correct `true` with a wrong `false`, destroying the state `_verify_physical_device` reads. There is no read-only variant; `get_device_cert_state` returns a model without persisting, but is a different shape and not a drop-in.

**3. Six callers read the file directly, and they are not doing the same thing.** Some report the record *as* a record, which is honest — `/proxy/cert/status` says "cached installation state" and points at `POST /cert/verify`. Others rendered it as current fact. Telling them apart required reading each one; nothing in the API distinguishes "what we recorded" from "what is true now".

## Why I am not proposing a design here

Two candidate directions, and I do not think the choice should be made from the bug reports alone:

- **Kind-specific verifiers behind an explicit interface**, so "verify this device" dispatches on all three kinds and the caller cannot get a simulator answer for a phone.
[...]

PR #166, opened 2026-09-14
Stop silent state loss: verification clobbering the record, and capture changes dropping processes

jerimiah797:
Two bugs in the same family — state disappearing with nothing said — found by
running the cert-consent work against a real simulator rather than reasoning
about it. Bundled because they are one theme and both are small.

## 1. Verification erased fields it had no opinion about

After an auto-install and an erase on a live simulator, `installed_at` read
`None` — the field the erase path's own docstring calls *"what tells a later
reader it had the CA"*.

`is_cert_installed` rebuilt the entry from the four things it had just learned
and `update_cert_state` replaced rather than merged. So **every verification
erased `installed_at`** — and since the cache was deleted in #158, every caller
verifies, so it happened almost immediately after any install.

It took **`wifi_proxy_configs`** with it: a physical device's recorded proxy host
and `client_ip`, which is what `_verify_physical_device` reads to find that
device's traffic at all. Latent today only because physical devices are scoped
out of that path — and one call site from live. This is **D8** in the spec, "a
read-shaped call writes to disk", with a measured consequence.

### The fix is two halves, because either alone is insufficient

- `update_cert_state` merges per field, so **omission preserves**
- the verification sites pass what they *learned*, not a full `model_dump()`
  — which names every field including the ones it has no opinion about, and
  naming a field is how you clear it

That distinction is the contract: **omission preserves, naming overwrites even
with `None`**. The erase path depends on the second half to withdraw a trust
claim, so there is a test for each direction.

### Verified live, not just in tests

Re-ran the same demo against the fix:

| | before | after |
|---|---|---|
| `installed_at` after install | `None` | `2026-09-14T16:43:54Z` |
| after erase | `cert_installed: False`, `installed_at: None` | `cert_installed: False`, **`installed_at` preserved** |

The record now reads as a pair — *this device had the CA at 16:43:54 and does
not now* — which is the distinction **D15** needs, surviving the erase.

### A design consequence, noted on #149

`installed_at` is written **only by `install_cert`**, so it means "when quern
installed it", not "when this device last trusted the CA". A CA installed by
hand with `simctl keychain add-root-cert` verifies as trusted with
`installed_at` still `None` — so D15 built on it would mis-classify exactly the
people who took resolution #1 in the 428 and installed it themselves. The field
that answers D15 is one written by *verification*: `last_trusted_at`. Recorded
on #149 rather than added here, since it belongs on `TrustClaim`.

## 2. Capture changes dropped processes silently

`set_local_capture` replaces the list, and the response echoes only the new one
— so removing a process looks identical to adding one. I hit this myself:
passing two of three processes during the demo deleted the third, with nothing
said in the response, the log, or `proxy_status`.

Worse on the likeliest path: `MobileSafari` and `com.apple.WebKit.Networking`
are defaults applied only when **nothing** is specified, so
`quern enable-local-capture Metatext` leaves Metatext alone in the list and
web-view traffic silently stops being captured. An agent told "capture
MobileSafari" does the same to whatever was there before.

Replacing is correct for `set`, so that is unchanged — the defect was that it
was invisible. The API logs what it dropped; the CLI prints it.

## Tests

Eight new. Six mutations caught:

| mutation | |
|---|---|
| restore the wholesale replace | 4 failed |
| verification writes a full `model_dump()` again | 3 failed |
| merge ignores explicit `None` (cannot clear) | 3 failed |
| merge leaks across devices | 1 failed |
| stop reporting drops | 2 failed |
| warn on every change, including additions | 1 failed |

That fourth one initially survived — the cross-device test wrote `aaa` first, so
a merge taking *any* device's record as its base still gave the right answer.
Rewritten so the device under update is written second and the first carries
fields it must not inherit.

2318 passing, 8 integration passing against real simulators, ruff clean.

## Follow-up

**#165** — installing an app on a proxied device should offer to add it to the
capture list. A wipe removes the app while the list survives, so capture ends up
naming a process that can never run; reinstalling is the moment to fix that, and
quern has both halves of the information and connects neither.


jerimiah797:
Merging with `scripts/merge-pr.sh --force`, i.e. deliberately skipping the review gate. Recording why, since that flag exists so the skip is a decision rather than a habit.

**CodeRabbit is rate-limited and did not review this head.** Its only comment is the limit notice, so the green `CodeRabbit` check means the bot ran, not that it looked. `pr-review-status.py` correctly reports `WAIT — pushed since the last review`.

**What did review it:** a local agent, against this exact diff. It found four code defects and six test gaps, all fixed in `4b7bbf0`:

- three install sites still wrote a full `model_dump()` and still clobbered `installed_at` / `wifi_proxy_configs` — including the Android already-installed branch, which is the bug this PR is named for on the sibling path
- a failed `list_devices` renamed the device on disk, because `_get_device_name` returned `"Unknown Device"` for both "absent" and "lookup failed"
- the merge preserved non-canonical fields forever, where the old wholesale replace had been healing them
- `_verify_physical_device` wrote a pre-`await` snapshot back over the record

and on tests: two of mine passed if verification wrote *nothing at all*; nothing asserted `verified_at` was refreshed; three warning tests asserted on message substrings rather than the record.

**Verification standing at merge:** all six CI checks green on 3.11–3.14, 2322 unit tests, 8 integration tests against real simulators, ruff clean, seven mutations caught, and the `installed_at` fix confirmed live on a real simulator rather than only in tests.

PR #167, opened 2026-09-14
Refresh the update check after an update, by asking rather than inferring

jerimiah797:
**Reported:** the menu bar kept offering an update that had already been
applied. Running `quern check-updates` fixed it immediately.

`~/.quern/update-info.json` holds a version and `update_available` computed
*before* the run, and `~/.quern/last-update-check` is a `touch()`ed marker that
suppresses a fresh check for 24 hours. Nothing in the update path touched
either. Three update runs left both untouched; one check run fixed it.

`invalidate_update_check()` already existed for exactly this — documented as
*"discard the cached update check so the next one runs immediately"* — and was
wired only to a channel change. Same shape as several bugs this week: the right
function exists and the caller was never connected to it.

## Two wrong fixes first, both caught by review

Worth recording, because both were **worse than the bug**:

**Inferring `update_available: false` from `rc == 2`.** `_update_via_git`
returns 2 from three places and only one means "at the channel tip". Another
prints *"`origin/release/stable` is N commits ahead, switch and rerun"* — where
`has_updates` is **true**. Writing "up to date" there hides a real update *and*
stamps the 24-hour limit so nothing corrects it for a day. The original bug
over-offered, which is annoying; this under-offered, which is silent. The menu
bar's own source calls a checkout ahead of its release branch "the common case".

**Deleting the files.** Also a false all-clear: `UpdateInfo` in the menu bar
defaults `updateAvailable` to false, so a *missing* file renders as "Up to
date". The up-front invalidation destroyed a true answer on paths where nothing
was even attempted.

## What it does now

On the paths where the run completed, a **real forced check** writes the truth.
No inference from a return code, correct for all three meanings of `rc == 2`,
correct on beta, and it goes through the existing locked writer.

- **Forced** skips the opt-out, deliberately and consistently with
  `check-updates`: you ran an update command, which is an explicit request, and
  that setting governs the *automatic* check. It is also what stops an
  opted-out user being left with no record.
- **Failures leave the previous record alone**, because absent is not neutral.
- **The lock and the race resolved themselves** with the redesign.
  `check_for_updates` writes under `_channel_lock()` and re-reads the channel
  inside it; removing the up-front invalidation also removed the second writer,
  since the restart's startup check stays rate-limited.

## Also fixed here

`_check_via_quern_dev` omitted `channel`, so quern.dev assumed stable and a beta
user at the stable pointer was told there was nothing to update to.
Pre-existing, and it matters more now that this answer is what `run_update` acts
on — `check_for_updates` documents the same hazard at its own call site.

## Tests

The key one is an invariant rather than a case: **`run_update` must never write
`update-info.json` on any path**, because only a real check may. Asserted across
rc 0, 1 and 2, and it is what makes the three meanings of rc 2 irrelevant here.

Six mutations caught: inferring up-to-date from rc 2, never refreshing,
refreshing without force, refreshing on a failed update, dropping the channel,
and letting a refresh failure fail the update.

2330 passing, 8 integration passing against real simulators, ruff clean.

## Not changed

The zero-byte `last-update-check` is intentional — `touch()`ed and read via
`stat().st_mtime`, so it carries its meaning in the mtime.

## Summary by CodeRabbit

* **Bug Fixes**
  * Update checks now respect the configured update channel, including beta channels.
  * The update menu now refreshes after completed updates, including runs where no update was needed, so already-applied updates are no longer shown.
  * Failed updates preserve the existing update status and do not refresh it.
  * Refresh errors no longer cause an otherwise successful update to fail.


jerimiah797:
Merging with `scripts/merge-pr.sh --force`, i.e. deliberately skipping the review gate. Recording why, since that flag exists so the skip is a decision rather than a habit.

**CodeRabbit is rate-limited and did not review this head.** The green `CodeRabbit` status means the bot ran, not that it looked.

**What did review it: two local agent rounds against this exact diff**, and both found defects serious enough that the fix was reworked rather than patched.

*Round one* rejected the whole approach. Inferring `update_available: false` from `rc == 2` is false whenever `_update_via_git` returns 2 because you are on a feature branch and the release branch is ahead — it prints "switch and rerun" there. That would have hidden a real update and stamped the 24-hour limit so nothing corrected it. Deleting the files instead is also a false all-clear, because `UpdateInfo.updateAvailable` defaults to false and a missing file renders as "Up to date". The fix now asks rather than infers.

*Round two* found two more. A refresh that could not reach the endpoint left the stale record in place **and** a fresh rate-limit stamp, turning an hour of staleness into a day — the reported bug, made worse by the fix for it. And the refresh was skipped on two exits where the pull had already succeeded, so a failed rebuild or tool upgrade left the menu bar offering a version that was installed.

It also verified the things I had taken on faith: `quern.dev` accepts `sha` + `channel` together, malformed channels fall back to stable without flipping the answer, and nothing can write a pre-rebuild version.

**Verification standing at merge:** six CI check runs green on 3.11–3.14, 2334 unit tests, 8 integration tests against real simulators, ruff clean, and mutation coverage on every behaviour the branch adds — including the no-inference invariant, which is what makes the three meanings of `rc == 2` irrelevant here.

Issue #184, opened 2026-09-15
No webinspectord socket is created under Xcode 27, so get_web_content has no transport


jerimiah797:
Observed on Xcode 27.0 (build 27A266a), macOS 26, Apple Silicon. **I have not established the cause** — this is the evidence, and a hypothesis at the end.

## Symptom

`get_web_content` fails on every simulator:

```
[web-content] could not reach the simulator's Web Inspector: no live webinspectord
socket among 1 candidate(s). Socket files persist after a simulator shuts down, so
most are stale. Tried: ['/private/tmp/com.apple.launchd.WMPOCYQKIr/com.apple.webinspectord_sim.socket:
[Errno 61] Connection refused']
```

The message is accurate. The problem is that the one candidate is genuinely stale, and nothing has replaced it.

## Evidence

```
$ ls -dlt /private/tmp/com.apple.launchd.*
drwx------  3 jham  wheel  96 Sep 14 09:31 /private/tmp/com.apple.launchd.ue16EbrQ9B
drwx------  3 jham  wheel  96 Sep 14 09:31 /private/tmp/com.apple.launchd.TocHWVWVoV
drwx------  3 jham  wheel  96 Sep 14 09:31 /private/tmp/com.apple.launchd.WMPOCYQKIr
```

- Every one of them predates the Xcode 27 upgrade. **No `com.apple.launchd.*` directory has been created since**, despite booting a simulator today.
- `find /private/tmp -maxdepth 3 -type s -newermt 2026-09-15` finds mitmproxy, idb and other sockets from today — but no webinspectord one. So socket creation in general is working; this one specifically is absent.
- Inside the booted simulator, the daemon is registered but not running:

```
$ xcrun simctl spawn <udid> launchctl list | grep -i webinspect
-	0	com.apple.webinspectord
```

## Ruled out

**The Safari Web Inspector setting.** This was my first hypothesis and it is wrong:

```
$ xcrun simctl spawn <udid> defaults write com.apple.mobilesafari \
    WebKitDeveloperExtrasEnabledPreferenceKey -bool true
```

No socket appeared, and `get_web_content` still failed. Worth recording because it is the obvious guess and it costs a detour.

[...]

Issue #170, opened 2026-09-15
A /source one second over the timeout destroys the WDA runner, making the device unautomatable


jerimiah797:
Measured on a physical iPhone 11 (iOS 26.6.2), Mac mini M4, device on a direct front port.

## The cascade

`describe_all` (`server/device/wda_client.py:841`) GETs `/source` with `self._source_timeout(udid)` — 3s, or 6s for older devices. On timeout it checks `_is_wda_responsive`, and if that fails calls **`_restart_wda`**, then returns `build_screen_skeleton`, which is empty.

On this device the first `/source` takes **7.52s**:

```
$ curl -s -m 30 -o /tmp/src.json "http://127.0.0.1:18100/source?format=json"
real 7.52
```

So the default path: times out at 6s → restarts WDA → returns an empty tree. And `_restart_wda` goes through `xcodebuild test-without-building`, which reinstalls the runner (#159) and is exactly the step that fails on an unhealthy install channel.

One second of latency therefore becomes a device that cannot be automated at all, and it is self-perpetuating: every subsequent call repeats it.

## Why it presents as "WDA is broken"

The symptoms actively point away from the cause:

- `start_driver` returns `{"status": "already_running", "ready": true}` — a forward exists, so it looks healthy
- `get_screen_summary` returns `element_count: 0` with **no error** — reads as an empty screen, not a failure
- `take_screenshot` works fine, because it is a different path — so the device looks alive
- WDA itself is perfectly healthy throughout: `/status` → `ready: true, state: success`, and `/source` returns a full tree with 331 elements

I spent a long time concluding WDA was broken, then that it was USB re-enumeration, then tunnel rotation (#162). All three were wrong. Ruled out by measurement:

- **Tunnel rotation:** the tunnel address was stable across 5 samples over 60s (`fdd9:e835:7449::1:51750` every time).
- **USB:** the device is on a direct front port; the Apple `USB2 Hub` above it is the Mac's internal controller hub.
- **WDA lifetime with quern uninvolved:** launched by hand, it stayed `ready: true` for 90s. Every death coincided with a quern call.

## Proof

Same device, same session, only the timeout differs:

| call | result |
|---|---|
| `get_screen_summary` (default) | `element_count: 0`, harness killed |
| `get_screen_summary` with `source_timeout: 25` | **`element_count: 331`** |

## Suggestions

[...]

PR #187, opened 2026-09-15
Build WebDriverAgent against a deployment target Xcode 27 accepts

jerimiah797:
Third Xcode 27 breakage, after #176 (SimulatorKit moved) and #184 (webinspectord socket). This one takes out physical-device automation entirely for anyone who has to build WDA.

## The failure

```
error: The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 13.0,
but the range of supported deployment target versions is 15.0 to 27.0.x.
(in target 'WebDriverAgentRunner' from project 'WebDriverAgent')
```

Upstream WebDriverAgent still declares 13.0; Xcode 27 rejects anything under 15.0. So `setup_wda` returns HTTP 500 and there is no WDA — no `get_ui_tree`, no taps, nothing — on any machine that has to build it.

## The fix

One build setting, passed on the command line rather than patched into the vendored project, which is a clone that gets re-pulled. 15.0 is Xcode 27's floor and far below anything quern supports on a physical device — tunneld needs iOS 17+ — so raising it costs nothing today.

## Why nobody noticed, and why it is nastier than it looks

`build_wda` skips when `build_team_id` matches. It tracks **nothing about the Xcode or SDK version**, so an existing build survives a major Xcode upgrade and `setup_wda` cheerfully reports `built: false`.

That means:

- anyone with a cached build keeps working, and has no idea this is broken
- anyone setting up fresh, adding a device, or forcing a rebuild is dead on arrival
- **the cache that hides the breakage is the only thing keeping existing users working** — and the first thing to force a rebuild deletes derived data *before* building (`if force and WDA_DERIVED.exists(): rmtree`), so it discovers the failure only after destroying the working build

I hit that last one on this machine: a forced rebuild wiped `quern-driver.xctestrun`, the build failed, and `start_driver` was left with nothing. Recovered through the no-install path in #159.

**Not fixed here.** Invalidating the WDA cache on an Xcode change is a separate change with its own argument, and bundling it would hide a behaviour change inside a build-flag fix. Worth its own issue.

## Verified on the device, not only in tests

| | before | after |
|---|---|---|
| `setup_wda force:true` | HTTP 500, xctestrun destroyed | `built: true, installed: true` in 28s |
| `start_driver` | `No .xctestrun file found` | `ready: true` |
| Phone takes input | — | `press`/`tap` 200 |

## Tests

Three, and three mutations caught: removing the override, restoring 13.0, and setting the floor absurdly high. The last one matters because a build flag could silently narrow the supported device range — so the test pins **both** ends, not just the one that fails today.

## One thing this does not fix

With the freshly built WDA, `/source` on the iPhone 11 now measures **10.19–10.35s**, against the 10.0s floor 0.18.0 set from a then-measured 7.52s. So the default budget is marginally too low again and the tree comes back empty — `source_timeout: 30` gives 328 elements.

That is #170's third bullet arriving from an angle it did not anticipate: the threshold moved because the *toolchain* changed, not the device or the screen. Deliberately not bundled here — it is a different fix with a different argument, and a second arbitrary constant does not belong inside a build-flag change.

## Summary by CodeRabbit

* **Compatibility**
  * WebDriverAgent builds now support a minimum iOS deployment target of 15.0, restoring physical-device builds with newer Xcode versions.
  * Deployment targets remain within the supported device range.

* **Bug Fixes**
  * Improved reliability when retrieving page source from both newer and older devices by allowing additional response time.


jerimiah797:
## Added the timeout raise (`d27c36a`)

Folding in the `/source` budget after all — 20s for A13-and-older, 10s default.

**The slow value is measured.** Same iPhone 11, same home screen, same quern; only WDA rebuilt under Xcode 27:

| WDA built with | `/source` | Budget then | Result |
|---|---|---|---|
| Xcode 26 | 7.52s | 6.0 → raised to 10.0 in 0.18.0 | fixed |
| Xcode 27 | 10.19 / 10.24 / 10.35s | 10.0 | **empty tree again** |

Doubled rather than nudged past the new number, because sizing the margin to the last observation is what produced this twice in one day. The test pins the **headroom** (≥1.5×) rather than the threshold, so a future value that merely clears the last measurement fails.

Verified on the device: the call that returned `element_count: 0` now returns **328 at the default, no override**, in 10.6s.

**The default is not measured, and I have left that visible.** No modern device has been timed under Xcode 27. 5.0 → 10.0 is precautionary — the toolchain shift that doubled the iPhone 11's time was not device-specific, so a modern device is *probably* slower too. Probably, not known.

A mutation reverting it to 5.0 **survives**, and I did not write a test to kill it. Pinning an absolute value there would manufacture evidence that does not exist; the constant says so in a comment instead. Worth measuring on an A16/A17 before anyone treats 10.0 as meaningful.

Mutations, otherwise all caught: reverting the slow budget to 10.0, setting it to 11.0 (above the measurement but with no headroom), and collapsing the two budgets into one.

**Still not the fix.** Deriving the budget from the first successful `/source` per device would delete both constants and end this cycle — the open half of #170. Two arbitrary constants raised twice in a day is the argument for doing it.

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #198, opened 2026-09-16
Spell the command the way the reader can actually type it

### LINKED ISSUE #192: CLI messages say "./quern", which is unnecessary for git installs and wrong for tarball ones
Error messages and check details across the CLI tell people to run `./quern doctor`, `./quern tunneld install`, `./quern setup` and so on. The `./` has not been necessary since `quern setup` began writing `~/.local/bin/quern`, and for a tarball install it is actively awkward — the reader is not necessarily standing in the extracted directory.

## Scope

| area | occurrences |
|---|---|
| `server/` | 42 |
| root `*.md` | 17 |
| `docs/` | 10 |
| `tests/` | 3 |
| `scripts/` | 2 |

Concentrated in `server/lifecycle/setup.py` (16) and `server/device/tunneld.py` (16), then `pmd3.py` (4), `updater.py` (2), `capture_env.py` (2), `daemon.py` (1), `idb.py` (1).

## Not a blanket replacement — there is a first-run case

`sed`-ing all 42 would introduce a bug. Some of these messages fire *before the wrapper exists*:

```
server/lifecycle/setup.py:2198   print("  Install it first, then re-run: ./quern setup")
server/lifecycle/setup.py:2742   report.add(install_wrapper_script())
```

`~/.local/bin/quern` is written at line 2742. Anything that bails before then — a missing prerequisite, a declined virtualenv prompt (#130), a setup with no terminal — is talking to someone who has no `quern` on `PATH` and got there by running `./quern setup` from a clone or an extracted tarball. Telling them to run `quern setup` names a command they do not have.

Most of `setup.py`'s messages are in this position, because the `check_*` functions all run before line 2742.

The same strings are *also* shown by `quern doctor` after a successful install, where `./quern` is exactly the wrong advice. So the text is correct in one context and wrong in the other, which is why a static choice cannot win.

## Three categories, three answers

1. **Post-install runtime messages** — `tunneld.py`, `pmd3.py`, `idb.py`, `daemon.py`, and anything reached from a running server. Setup has demonstrably succeeded, so: **`quern`**. This is the bulk of the complaint and the easy win.

2. **Messages reachable before `install_wrapper_script()`** — most of `setup.py`. Decide at runtime rather than in the source:

   ```python
   def _quern_cmd() -> str:
       """How to invoke quern, from where the reader is standing."""
       return "quern" if shutil.which("quern") else "./quern"
   ```

   One helper, and the message is right in both contexts. `WRAPPER_PATH.exists()` is the weaker test — the wrapper can exist while `~/.local/bin` is not on `PATH`, which `CONTRIBUTING.md` already documents as a live failure mode (zsh caches the first resolution, so `type -a` and what actually runs can disagree).

3. **Prose naming the wrapper script itself** — `capture_env.py:152,157` explain that `./quern setup` re-execs into `<project>/.venv`, and that "anything that bypasses the `./quern` wrapper gets the checkout's answer". That is describing a file in the repo, not instructing anyone. **Leave it.** Same for `setup.py:586`, the provenance comment written into the generated wrapper.

## Docs

`README.md` and `CONTRIBUTING.md` first-run instructions must keep `./quern setup` — on a fresh clone or a freshly extracted tarball there is no other way to start, and that is true for both install kinds. Everything past the install step can drop the `./`.

## Notes for whoever takes it

- `tests/` has 3 occurrences; check whether any assert on the literal string, and whether `tests/test_readme_sync.py` compares documented commands against CLI dispatch.
- Coordinate with #191, which already touches `setup.py`, `pmd3.py` and `idb.py` (21 of the 42). Branch after it merges or expect a rebase.
- Worth adding a test that no *post-install* message contains `./quern`, or this reaccumulates — every new check detail is a fresh chance to type it.

jerimiah797:
Closes #192.

Error messages and check details told people to run `./quern doctor`, `./quern tunneld install`, `./quern setup`. The `./` has not been necessary since `quern setup` began writing `~/.local/bin/quern`, and it is wrong for a tarball install — the install script leaves quern in `~/.local/share/quern` and your terminal wherever it was.

## Why this isn't a `sed`

The same strings print in situations with different right answers:

- `quern doctor` runs **after** install — `./quern` is merely awkward.
- Most of `setup.py`'s checks run **before** `install_wrapper_script()` — `quern setup` names a command the reader does not have.
- And in neither case is the reader necessarily standing in the project directory, which is what `./` assumes.

## Three answers

`server/config.quern_cmd()` resolves per call:

| situation | answer |
|---|---|
| wrapper on PATH | `quern` |
| no wrapper, cwd **is** the project | `./quern` |
| no wrapper, cwd is elsewhere | absolute path, `shlex.quote`d |

`shutil.which` rather than `WRAPPER_PATH.exists()` — the wrapper can exist while `~/.local/bin` is off `PATH`, which `CONTRIBUTING.md` documents as a live failure.

**Not cached.** The answer depends on `PATH`, on the wrapper existing, *and* on the current directory, and all three move within a process: setup installs the wrapper partway through its own run, uninstall removes it, anything may `chdir`. An earlier revision cached it and was wrong the moment the answer became cwd-dependent.

## The wrapper had to change too

`quern_cmd()` now hands people an absolute path, and the wrapper could not honour it: the pre-venv fallback `exec python3 -m server` never put the install on the import path. Its comment recorded the assumption — *"still works from the clone and is where setup is run from"* — which is the assumption this PR exists to challenge.

Fixed with `PYTHONPATH` **and `-P`**. The `-P` is load-bearing, not belt-and-braces: `-m` puts the caller's cwd at `sys.path[0]`, *ahead* of `PYTHONPATH`, and quern's users are by design standing in their own app repositories where `server/` is an ordinary directory name. Without it, theirs shadows ours:

```
No module named server.__main__; 'server' is a package and cannot be directly executed
```

`-P` drops cwd from the path while leaving the process there, so relative arguments — a project directory to build — still resolve against it. That is why this is not `cd "$root"`.

## The guard

A test AST-scans every module and fails on a hardcoded `./quern` in any non-docstring literal, because otherwise this reaccumulates.

Exemptions are **literal-precise, not file-level**. An earlier revision allow-listed `setup.py` wholesale to cover one f-string template, which exempted the file with the most message sites — a hardcoded `./quern doctor` anywhere in it passed. The decoy test now runs the real scanner over a planted tree; re-implementing the walk inline proved `ast` works and nothing about the guard.

Known limit, accepted: `"./" + "quern"` or a `%`-format defeats it. Building that string in pieces is not an accident anyone has by mistake.

## Docs describe both installs

Framed as two *moments*, not two audiences: during first setup there is nothing on `PATH`, so you run it by path — `./quern setup` in the directory, the full path if you are not. Afterwards `quern` works from anywhere. README and getting-started both say so, including the `rehash` case. Clone blocks keep `./quern` and say why: with several checkouts the bare name resolves to whichever setup last pointed at.

Five `docs/guides/*` pages need the quern.dev sync at the next release; `sync-docs.py --check` names them. That belongs to the release doc pass, not this branch.

## Review history

Both review rounds found real defects, several introduced by this branch. Worth knowing what they were:

- **The highest-value message was not converted.** `run_it_yourself(["quern", "setup"])` prints a bare `quern setup` on the declined-venv exit — which its own docstring says a `curl | bash` setup hits *every time*. It appeared next to a correctly-converted line in my own live-test output.
- **`-P` was missing**, so the wrapper fix did not do what its comment claimed.
- **Two tests proved nothing** — the decoy re-implemented the scanner; the absolute-path test asserted `endswith("/quern")` and `exists()`, both satisfied by the project *directory* in any checkout named `quern`, including CI.
- **The wrapper had zero coverage** — reverting it passed all 2490 tests, since CI always builds a venv and never takes the pre-venv branch. Two subprocess tests cover it now.

2496 pass, lint clean.

## Summary by CodeRabbit

- **New Features**
  - Quern can now be run reliably from any directory after setup, including installations where the command is available on `PATH`.
  - Command guidance adapts to the way Quern is installed and invoked, including setup, tunneling, crash recovery, and troubleshooting instructions.
  - Improved wrapper behavior prevents similarly named local packages from interfering with Quern.

- **Documentation**
  - Updated installation, getting-started, device, iOS, Android, and troubleshooting examples with clearer command usage and recovery guidance.
  - Added instructions for resolving shell `command not found` messages.


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #199, opened 2026-09-16
Clear Android text fields through the IME, and check that they cleared

### LINKED ISSUE #177: clear_text on Android deletes one character per call and reports ok
## What happens

`POST /api/v1/device/ui/clear` on Android deletes **exactly one character** per
call and returns `{"status": "ok"}` every time.

Measured against a Pixel 3 XL (Android 10), server v0.17.0, on a field holding a
known string:

```
after type : 'conformancab_CD!2@xABCDEFGHIJxyzkeep-meto-be-clear'
clear #1 -> HTTP 200 {"status":"ok"}   value now: '…to-be-clea'
clear #2 -> HTTP 200 {"status":"ok"}   value now: '…to-be-cle'
clear #3 -> HTTP 200 {"status":"ok"}   value now: '…to-be-cl'
```

## This is #98, on the other backend

[#98](https://github.com/quern-dev/quern/issues/98) was the same failure:
select-then-single-backspace, where the selection silently does not take, so one
character goes and the call reports `ok`. That was fixed for the web path in
`controller_ui.clear_text` — which now verifies via `_clear_web_input` and
*refuses* with `_MAX_DELETE` rather than half-clearing.

The native path in the same function got neither guard, and the Android backend
has the identical construction. So the branch with no verification at all is the
one still carrying the bug.

## Why

`server/device/u2_client.py:541`, `select_all_and_delete`:

```python
device = self._connect(udid)
device.click(int(x), int(y))     # u2 used here…
import subprocess                # …and then not for the part that matters
subprocess.run([... "input", "keyevent", "KEYCODE_MOVE_HOME"], ...)
subprocess.run([... "input", "keyevent",
                "--longpress", "KEYCODE_SHIFT_LEFT", "KEYCODE_MOVE_END"], ...)
subprocess.run([... "input", "keyevent", "KEYCODE_DEL"], ...)
```

`input keyevent` given several keycodes sends them **in sequence, not as a
chord**. Shift cannot be held across separate keycodes this way, and
`--longpress` applies to the sequence rather than making Shift a modifier. So
nothing is ever selected: the caret goes to the start, then to the end, and the
lone `KEYCODE_DEL` backspaces once. The observed behaviour is exactly what the
code does.

### It is not a missing IME

Checked, because that is the obvious first theory:

```
$ adb shell settings get secure default_input_method
com.github.uiautomator/.AdbKeyboard          ← installed AND active
$ python -c "import importlib.metadata as m; print(m.version('uiautomator2'))"
3.7.0
```

The companion app is installed and its keyboard is the active input method —
which is why **typing works**, including the shift-character string `ab_CD!2@x`,
which arrives verbatim. The defect is only that `select_all_and_delete` abandons
the u2 handle opened on the line above and shells out to `adb` instead.

## Two further problems in the same nine lines

- **No return code is checked.** All three `subprocess.run` calls pass
  `capture_output=True` and discard the result, so every one could fail and the
  function would still return normally.
- **`adb` is taken from `PATH`** rather than the resolved binary the rest of the
  backend uses.

## Why it matters

Clearing a field is *setup*, not an assertion — it runs before the interesting
part. A clear that silently leaves 49 of 50 characters means the next
`type_text` appends, and the failure surfaces later as a mismatched string in an
unrelated assertion. That is exactly how it presented: three conformance tests
failed, only one of which was about clearing.

It also makes `clear_text` unusable as a reset between steps on Android, which
is the main thing it is for.

## Suggested fix

uiautomator2 3.7.0 exposes `Device.clear_text()` (confirmed present), and
`_connect(udid)` on the line above already holds the handle. `set_text("")` on
the resolved element works too. Failing that, `input keycombination
KEYCODE_CTRL_LEFT KEYCODE_A` (API 31+) sends a real chord.

Whichever is chosen: **read the value back and raise when it is not empty.**
That is the guard #98 added to the web path, and its absence here is why this
went unnoticed.

## Guard

`tests/conformance/test_ui.py::test_clearing_a_named_field_empties_it[android]`
on `test/api-conformance-suite`, currently failing against this bug.

jerimiah797:
Fixes #177.

`clear_text` on Android removed **exactly one character per call** and returned
`{"status": "ok"}`. Measured on a Pixel 3 XL — 36 characters in, 35 out, success
reported.

## Cause

`select_all_and_delete` opened a uiautomator2 connection, used it only for the
focusing tap, then shelled out:

```
input keyevent KEYCODE_MOVE_HOME
input keyevent --longpress KEYCODE_SHIFT_LEFT KEYCODE_MOVE_END
input keyevent KEYCODE_DEL
```

`input keyevent` given several keycodes sends them **in sequence, not as a
chord**, so Shift was never held and nothing was selected: caret to the start,
caret to the end, one backspace. All three calls also discarded their return
codes.

Not an IME problem — `AdbKeyboard` was installed and active throughout, which is
why typing worked. The fix uses `device.clear_text()`, which broadcasts
`ADB_KEYBOARD_CLEAR_TEXT` to that same IME and clears in one operation.

## The verification is the other half

#98 fixed this defect on the iOS **web** path and hardened it with a read-back;
the **native** path got neither, which is why this survived. Clearing is setup,
so a partial clear surfaces later as a mismatched string somewhere unrelated —
it failed three conformance tests of which only one was about clearing.

**Live testing caught my first attempt at the check.** An empty Android
`EditText` reports its *hint* in `text`, and uiautomator exposes no separate
hint attribute:

```
empty field   → text: 'default'   (the hint)
after typing  → text: 'HELLO'
after clear   → text: 'default'
```

So "the field must read empty" fails on every hinted field on the platform,
against fields cleared correctly. What the bug actually leaves is a *shortened
piece of the original*, so the check compares before against after and fires
only when what remains is a fragment of what was there. It stays quiet when the
two are equal, which is ambiguous between "already empty" and "did nothing" —
better than a false alarm on a legitimate repeat clear.

## Verification

**Live, both targets** — Pixel 3 XL (Android 10) and Pixel_7 emulator (API 36):

| | before | after |
|---|---|---|
| 36 chars, one clear | 35 chars left, `200 ok` | **empty**, `200 ok` |
| second clear | 34 chars left | idempotent |
| neighbouring field | — | untouched |

**Conformance suite (Android):** `21 passed, 0 failed`.
`test_clearing_a_named_field_empties_it[android]` has failed on every run this
session and now passes; reverting this change makes it fail again.

**iOS re-run for collateral damage:** `24 passed, 1 failed` — the failure is
#84, pre-existing and unrelated (this branch does not carry that fix).

**Unit suite:** `2489 passed`.

**Mutation-tested**, from a `git archive` copy:
- dropping the read-back → `test_a_partial_clear_raises_rather_than_reporting_ok` fails, alone
- restoring the `KEYCODE_DEL` subprocess → `test_clears_through_the_ime_not_keystrokes` and `test_focuses_the_field_before_clearing` fail

## Also in this PR

The conformance test made the same wrong assumption the implementation did — it
asserted the field reads empty, allowing only the *identifier* as a placeholder,
which quietly covered iOS (`field_default`) and failed on Android (`default`).
It now asserts the typed content is gone, and separately that what remains is
not a fragment of it. That change is on `test/api-conformance-suite`, not here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved clearing of text fields on Android devices for more reliable deletion.
  * Detects when only part of a field’s content is removed and reports the issue instead of continuing silently.
  * Handles empty, unreadable, or hint-text field values more safely.
  * Preserves connection-related errors with clearer failure reporting.


jerimiah797:
## Review findings addressed — one was a real bug that would have shipped

**1. A correct clear raised when the hint is a substring of the content — fixed.**

Confirmed live before changing anything. `field_email` is hinted `email`;
clearing `email@example.com` out of it leaves `email`, a substring of the
original:

```
clear -> HTTP 500
{"detail":"[u2] clear_text left 5 of 17 character(s) in the field: 'email'"}
field now: 'email'      ← correctly emptied
```

An email address in an email field is the ordinary case. I had considered this
shape while writing the check and talked myself out of it as contrived, which
was wrong.

The substring test is replaced by an **idempotence** test, which is what
actually separates the two: a correct clear does not move when repeated, while
#177's bug removed one character per call and so shrinks the field again. Only
the ambiguous case — a non-empty result that is a substring of the original —
pays the extra round trip; a hint that is not a substring costs nothing.

Re-verified live on both targets after the change:

```
EMULATOR    email field  -> HTTP 200  value='email'
EMULATOR    plain field  -> HTTP 200  value='default'
PIXEL 3 XL  email field  -> HTTP 200  value='email'
PIXEL 3 XL  plain field  -> HTTP 200  value='default'
```

Regression test added (`test_a_hint_that_is_a_substring_of_the_content_is_not_a_failure`).
Android conformance 21 passed; unit suite 2490.

**2. Blind to a clear that removes nothing (`after == before`) — acknowledged, not fixed.**

Correct, and it was a deliberate choice I should have written down. The two
readings are genuinely indistinguishable: a field that was already empty and a
field whose clear did nothing both report the hint unchanged, because Android
exposes no way to tell an empty field from one containing the hint. Raising
would break the legitimate case of clearing an already-empty field, which is
ordinary setup.

Closing it properly needs a sentinel — type a character, clear, confirm it went
— which is more invasive than a clear should be. I would rather leave the gap
stated than add a false-alarm path. Worth noting the specific mechanism
suggested (an IME switch leaving no bound input connection) would also break
`type_text`, which uses the same IME and demonstrably works; so if it occurs it
should surface more loudly than here.

**3. `clear_text()` may fall back to `jsonrpc.clearInputText()` — fair, and the untested path is stated.**

Accurate: u2 branches on `is_input_ime_installed()`, and the mocked
`assert_called_once` cannot tell the branches apart. Every live measurement here
went through the IME path, since `AdbKeyboard` was installed and active. The
fallback is uiautomator's own `ACTION_SET_TEXT` route rather than a keystroke
loop, so it is not prone to #177's failure — but it is unexercised by this
change, and that is now said rather than implied.

jerimiah797:
@coderabbitai review

jerimiah797:
## Verification status, corrected

The body above says "live-verified on both" — that was true of the **first**
implementation. The check has since been rewritten twice, and the current one
(`871850f`, comparing `text` against the field's `hint`) is verified on the
**emulator only**:

- **Pixel_7 emulator (API 36)** — email address in a hinted email field, plain
  field, and clearing an already-empty field: all `200`. Android conformance
  21 passed. Unit suite 2491.
- **Pixel 3 XL (Android 10)** — **not verified against this implementation.**
  The device dropped off adb mid-session and is not physically reachable right
  now. The earlier physical run exercised a different check.

Recording it rather than letting the earlier result stand in for it.

## CI note

`test (3.14)` fails here, and identically on `main`:

```
tests/test_tool_probe.py::TestTheBackendIsNotLatchedAtStartup
  test_a_fresh_answer_is_not_re_probed  — "a cached answer was re-probed"
  test_a_stale_answer_is_re_probed      — "the cached answer was reused past its age"
```

Pre-existing, 3.14 only, in a module this PR does not touch (it changes
`server/device/u2_client.py` and `tests/test_u2_backend.py` only). The last two
runs on `main` fail the same way, including the #194 merge.

Issue #189, opened 2026-09-15
WDA build cache ignores the toolchain, so an Xcode upgrade never triggers a rebuild


jerimiah797:
`build_wda` keys its cache on the signing team and nothing else:

```python
if not force and state.get("build_team_id") == team_id:
```
`server/device/wda.py:304-305`

No Xcode version, no SDK version, no build settings. So upgrading Xcode — the event most likely to make an existing WebDriverAgent build wrong — is invisible to the cache, and quern goes on using artifacts compiled by a toolchain that is no longer installed.

Unlike #188 this one is usually benign: the stale artifact still runs, which is precisely why nobody notices until a rebuild is forced for an unrelated reason and it fails. Xcode 27 is the live example (#187): every machine that had built WDA before the upgrade kept working, and the breakage only appeared on the next forced rebuild — which is also the moment the old artifact gets deleted.

### The related gap #187 adds

`WDA_MIN_DEPLOYMENT_TARGET` (`server/device/wda.py:42`) is now an input to the build that the cache key does not cover. When Apple next moves the floor and that constant is bumped, existing installs will not rebuild. Benign for the same reason as above — the built artifact keeps working and nothing re-validates the setting — but it is one more instance of the same shape.

### Fix

Record the toolchain and the settings that affect the output alongside the team, and treat a mismatch as a miss:

```python
state["build_team_id"] = team_id
state["build_xcode_build"] = ...   # xcodebuild -version | DEVELOPER_DIR
state["build_deployment_target"] = WDA_MIN_DEPLOYMENT_TARGET
```

`xcodebuild -version` gives both the version and the build number; the build number is the one to store, since it changes for betas that keep the same marketing version.

Worth pairing with #188 — that one is the sharper bug (cache hit with no artifact at all), and a rebuild triggered here is only safe once a failed rebuild stops wedging the install.

jerimiah797:
#202 narrows this rather than closing it, and it is worth being precise about the gap so nobody assumes it is gone.

**What #202 does:** records the Xcode build id and deployment target alongside `build_team_id`, and rebuilds when a recorded value differs. An absent value is treated as "no opinion" rather than a mismatch, so existing installs are not all forced to rebuild on upgrade.

**What it does not do:** the cache is consulted from exactly one place.

```
_build_is_current  <- build_wda  <- setup_wda   (and nothing else)
```

[...]

PR #202, opened 2026-09-16
Ask the WDA cache all three questions, and say what a build failure means

### LINKED ISSUE #188: A failed force rebuild wedges WDA: the build is skipped, then install says "build first"
`build_wda` caches on the signing team alone and never checks that the artifacts it is caching actually exist:

```python
state = read_wda_state()
if not force and state.get("build_team_id") == team_id:
    logger.info("WDA already built for team %s", team_id)
    return False
```
`server/device/wda.py:304-307`

`build_team_id` is written only after a successful build (`server/device/wda.py:400`), but `force` removes the derived data *before* building (`server/device/wda.py:311-314`). So a force rebuild that fails leaves the state from the last *successful* build in place with nothing on disk to match it.

### The loop

1. A build succeeds. `build_team_id = TEAM`, artifacts present.
2. A forced rebuild runs — which `setup_wda`'s free-account warning and `docs/guides/ios-wda.md` tell users to do **weekly** — and fails. Derived data is already gone.
3. `setup_wda` without `force` now skips: *"WDA already built for team TEAM"*.
4. `_runner_app_signature_valid()` returns False, `_post_process_runner_app` logs *"Runner app not found — skipping"*, and `install_wda` raises **"WDA app not found — build first"** (`server/device/wda.py:869`).
5. Building is exactly what step 3 refuses to do. Without knowing to pass `force`, there is no way out.

The error names the right action and the tool then declines to take it, while the state file says the work is already done.

### Why it surfaced now

Xcode 27 made step 2 fail for everyone (#187), so the trap went from rare to the default experience on an upgraded machine. It is independent of that fix, though — any failed forced rebuild does it, and #187 only removes the most common cause.

Worth noting this is *not* the same as the Xcode-version cache staleness: that one leaves a working stale artifact, which is benign. This one leaves nothing while reporting success.

### Fix

Gate the skip on the artifacts as well as the team, so a cache entry with no artifact is not a cache hit:

```python
artifacts_present = WDA_APP.exists() and XCTESTRUN.exists()
if not force and artifacts_present and state.get("build_team_id") == team_id:
```

Clearing `build_team_id` on every build failure path would also work and is the convention `CONTRIBUTING.md` already states — *"a success marker must not survive a failure"* — but checking the artifact is the stronger of the two, because it also covers the case where something outside quern removes the build.

jerimiah797:
Closes #188. Addresses #189 — see *Scope* below for what it does not cover.

Started as the build cache and grew, because the cache fix makes rebuilds more likely and a rebuild is not free for everyone. Four commits, one thread: **quern could not tell whether the WDA artifact on disk was still good, and when a build failed it could not say why in terms the reader could act on.**

## 1. The cache asked one of three questions

It was keyed on the signing team alone.

**#188 — the artifacts were never checked.** `force` removes the derived data *before* building, and `build_team_id` is written only after a build that returned zero:

```
state:  build_team_id = TEAM     <- from the last successful build
disk:   (nothing)                <- force deleted it, then the build failed
```

Every later run then skips the build, and `install_wda` raises **"WDA app not found — build first"** — which is exactly what the skip refuses to do. Without knowing to pass `force` there is no way out. I spent a session in that loop before filing it.

**#189 — the toolchain was not recorded.** Upgrading Xcode is the event most likely to invalidate the artifact and the cache could not see it. Xcode 27 is the live example: machines that had built WDA beforehand kept the old one and found out at the next forced rebuild — which is also the moment the old one is deleted. That is where the two bugs meet, and why the order matters: **#189 causes more rebuilds; #188 is what makes a failed rebuild survivable.**

`WDA_MIN_DEPLOYMENT_TARGET` joins the fingerprint too — #187 added it as a build input and its own review noted the cache could not see it.

**A missing key is not a mismatch.** Every existing install has no `build_xcode`, and treating that as stale would rebuild WDA for everyone on upgrade to detect a staleness we cannot confirm. Absent means no opinion; present-and-different means rebuild. Same for a toolchain that cannot be read — a probe that failed is not a toolchain that changed.

## 2. Build failures did not say what went wrong

`_RUNNER_FAILURE_PATTERNS` already translates the failures that matter most — free-account limits, expired profiles, a missing certificate — but it was reachable only from `_diagnose_runner_failure`, which reads a runner *log*. A build failing the same way printed twenty raw lines of xcodebuild.

That is backwards: signing is what consumes a slot, so the **build** is where a free account meets the limit first, and the build is where the message was worst. Both paths share one lookup now. It matters more with the toolchain in the cache key, because that rebuild is the right call and its failure still has to be legible.

## 3. Two Apple limits, merged into one sentence, with the wrong remedy

Making the table reachable from the build path put text in front of people that was rarely read — and it was wrong. Both the failure message and the setup warning said:

> Free Apple developer account limit reached (max 3 app IDs per 7 days). Wait for old app IDs to expire…

There are two limits, and they are not the same budget:

| budget | limit | WDA's share | clears by waiting? |
|---|---|---|---|
| App IDs registered | 10 per rolling 7 days | 2 (`dev.quern.driver`, `.xctrunner`) | yes |
| free-signed apps installed on a device | 3 at once | **1** — the runner | **no**, delete an app |

The error we match — *"The maximum number of apps for free development profiles has been reached"* — is the **device install** limit. So the number was right, the description was wrong, and the advice sent someone to wait out a week for a condition waiting does not affect.

Confirmed against a real device: `devicectl` reports one app, `QuernDriver / dev.quern.driver.xctrunner`. The `.xctest` ships *inside* the runner, so it is one installed app and not two — the old warning overstated quern's cost while understating the consequence.

Which gives the number worth saying at setup rather than leaving someone to infer it from a failure on a different app: **a free account has three slots and quern takes one for as long as it is used, leaving two.** Both messages now also name the trap that makes this fire early — Xcode counts *offloaded* apps toward the three, including Apple's own, so the device can look far emptier than the error suggests.

## Scope: what this does not fix

Raised in review, and the PR previously overstated it. The cache is consulted from exactly one place:

```
_build_is_current  <- build_wda  <- setup_wda   (and nothing else)
```

`start_driver` goes straight to `_find_xctestrun()`, and `install_wda` reads `WDA_APP` directly. Those are the everyday paths — including the auto-start behind an ordinary `tap` on a device where WDA is not running. So a stale artifact is still *used* exactly as before; what changes is that `setup_wda` now notices and rebuilds.

Combined with "absent means no opinion", that means for an existing install the toolchain check is a no-op until some unrelated event records a fingerprint, and even then it only fires when someone explicitly runs setup. #189's stated symptom — "quern goes on using artifacts compiled by a toolchain that is no longer installed" — is therefore narrowed rather than removed, which is why this says *Addresses* rather than *Closes*.

Detecting it at *use* rather than at setup is a larger change and belongs with #159, which needs the same "does the thing on the device match what we built" question answered for its install-skip.

## Verification

**Against the real install**, not only fixtures:

```
real state + real artifacts   -> current: True    (no surprise rebuild)
app deleted                   -> current: False   (the wedge, caught)
toolchain this machine records: 27A266a
```

The real state here has no `build_xcode`, so it exercises the migration path and correctly does not rebuild.

**Six mutations on the cache, all caught.** Two survived the first pass — deleting the fingerprint write, and recording an unreadable toolchain as `None` — because the comparison was tested and the *recording* was not. That is the gap a review caught in #194, reproduced here in a class I had named `TestTheFingerprintIsRecorded` that tested only the parser.

The message corrections are pinned by tests as well, because their failure mode is a plausible sentence rather than a crash: the old wording cannot come back, and the remedy has to be the one that works.

## For the reviewer

- `tests/test_wda.py` patches `server.device.wda.asyncio.create_subprocess_exec`, which is the **global** module object, so the patch reaches anything else that goes through it. The new toolchain probe runs after the build and was overwriting the recorded argv with `xcodebuild -version` — which is how four existing tests started failing. The helper now keeps every invocation and selects the build. **Worth checking whether anything else in the suite has the same exposure.**
- Follow-on work is filed rather than smuggled in: **#203** (read the embedded profile to detect expiry, device coverage and a vanished identity, instead of telling free users to rebuild weekly) and a note on **#159** explaining that this PR's fingerprint is most of what its install-skip check will need.

2529 pass, lint clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved WebDriverAgent build validation when required artifacts are missing or the Xcode toolchain changes.
  - Prevented stale toolchain information from being retained after an unsuccessful Xcode version check.
  - Added clearer diagnostics for common signing and provisioning failures, including account and app-limit issues.
  - Clarified the distinction between App ID registration limits and free-signed apps installed on a device.

- **Documentation**
  - Updated the iOS guide with revised free-account limits, app-slot usage, and recommended remedies for signing errors.


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #213, opened 2026-09-17
Finish an update in a process that has loaded only the new code

### LINKED ISSUE #212: Updating to 0.18.3 crashes after the pull: the old process imports the new setup.py
## What happened

Updating a git install from 0.18.2 to 0.18.3 (field report):

```
Updated successfully to v0.18.3 (36 commits).
Installing Python dependencies ... up to date.
Building MCP server... MCP server built successfully
Traceback (most recent call last):
  ...
  File ".../server/lifecycle/updater.py", line 591, in _rebuild_and_restart
  File ".../server/lifecycle/setup.py", line 24, in <module>
    from server.config import CONFIG_DIR, quern_cmd
ImportError: cannot import name 'quern_cmd' from 'server.config'
```

Setup never ran and the server was not restarted. `quern setup` by hand recovered it.

## Why

`quern update` pulls (git) or swaps (tarball) the source tree **inside the process that is running the old code**, and then carries on importing from it. `_rebuild_and_restart` imports `server.lifecycle.setup` for the first time *after* the swap, so it loads the **new** `setup.py` — which imports `quern_cmd`, added in 0.18.3 (#198), from `server.config`, a module the **old** process loaded at startup and still holds.

Reproduced: load `server.__main__`, `server.config` and `server.lifecycle.updater` from a v0.18.2 tree, replace the tree with v0.18.3, import `server.lifecycle.setup` → exactly this `ImportError`. Reloading `server.config` is the only change needed for the import to succeed; nothing else in the path is stale.

The traceback itself shows the mix: its frames are old line numbers printed against new source.

## Who is affected

Every update **to 0.18.3 from 0.18.2 or earlier**, git and tarball alike — the tarball path calls the same `_rebuild_and_restart` after swapping the directory. The update reports failure after the source has already moved, leaving setup and the restart undone.

Recovery: `quern setup`, then `quern restart`.

It was not caught because the release verification resolved the channels and inspected the tarball, but did not run an update *from* the previous version.

## Fix (0.18.4)

1. **Run everything after the swap in a fresh process.** The updater should hand off to the new code (`python -m server …`) instead of importing it into a process that loaded the old one. This protects every update *from* 0.18.4 onward.
2. **Tolerate a stale process in the meantime.** Users still on ≤0.18.2 will update straight to 0.18.4 using their old updater, which will import 0.18.4's `setup.py` exactly as above. That file therefore has to work against the old in-memory `server.config` — reloading it when it is stale.
3. **Test the handoff by updating from the previous release**, not only by checking what the channel resolves to.

jerimiah797:
Closes #212. The fix for 0.18.4.

## The failure

Every update **to** 0.18.3 from an earlier release ended like this, git and tarball alike:

```
Updated successfully to v0.18.3 (36 commits).
...
  File ".../server/lifecycle/setup.py", line 24, in <module>
    from server.config import CONFIG_DIR, quern_cmd
ImportError: cannot import name 'quern_cmd' from 'server.config'
```

The update was already applied, but setup and the restart did not run. `quern update` swapped the tree and carried on in the **same interpreter**. `server.config` had been imported before the swap, so it was the old copy; `setup.py` was imported after, so it was the new copy, and it asked the old module for a name added in 0.18.3.

## The fix, in two halves

**1. Updates from this release hand off.** After the pull or swap, `run_update` runs `python -m server update --finish` (the venv's interpreter, cwd = the project) and returns its result. The rebuild, setup, restart and result record all happen in a process that has only ever loaded new code. The parent flushes its output first. Before that was added, the live run printed "Updated successfully" *after* the child's whole setup summary whenever the output was piped, which is how the menu bar reads it. If the child dies before recording an outcome, the parent records a failure and prints `quern setup` / `quern restart`.

**2. Updaters from ≤0.18.3 can still install this release.** They will import this release's files the old way, and nothing in this release can change that. So `server/lifecycle/stale_modules.py` recognises such a process: `server.lifecycle.updater` is loaded but has no `finish_update`. In that case it reloads the three modules those updaters load before the swap. `setup.py` and `__main__.py` call it before their server imports. Outside that situation it does nothing. The module list comes from importing the updater of 0.18.1, 0.18.2 and 0.18.3; all three modules import only the standard library apart from `server.config`.

## Live test: real git, real updater, no mocks

A sandbox with a bare repo standing in for GitHub, an install with its own venv, `HOME` and `QUERN_STATE_DIR`, and stub `osascript`/`open`/`sudo`/`launchctl`/`pkill`/`killall` first on PATH:

| run | installed | pulls | result |
|---|---|---|---|
| control | 0.18.2 | v0.18.3 | **`ImportError: quern_cmd`**, same as the field report |
| 1 | 0.18.2 | this branch | exit 0, setup ran, record `updated` |
| 2 | this branch | a probe commit adding `HANDOFF_PROBE` to `server/config.py` and importing it at the top of `setup.py`, the #212 shape | exit 0, setup ran, output in order |
| control | this branch, hand-off disabled | the same probe | **`ImportError: HANDOFF_PROBE`** |

No stub was triggered, and the real wrapper, menu-bar app and server were untouched.

## Tests

- `tests/test_update_handoff.py`, 15 tests: the command, interpreter and cwd of the hand-off; `--tools` passed through; failure and missing-record handling; flushing before the child runs; recognising an old updater; reloading a stale module. Plus **`test_an_old_updater_can_import_this_release[v0.18.2, v0.18.3]`**, which loads the real old updater from a `git archive` of the tag, copies this tree over it, and performs the imports `_rebuild_and_restart` does, in the same order. It needs the tags, so it skips on a shallow checkout rather than faking them.
- Mutations: removing the reload fails 2 tests; removing the hand-off fails 8.
- The argv test in `test_tool_updates.py` used to assert on the source text of `__main__.py`. It now runs `main()` for all four flag combinations.
- Full suite: 2559 passed; ruff clean.

## What went wrong while writing it

**The first version of these tests ran a real `quern update --finish` in this checkout.** Existing updater tests call `run_update`, which now spawns a process. The child inherited the tests' `QUERN_STATE_DIR` but the real `HOME`, so setup rewrote the developer's `~/.claude/settings.json` hook to point at a temporary directory. Later, a mutation that removed the hand-off did the same thing in-process. Both were restored. The conftest path guard reported the first one, but only after the write.

So `conftest.py` now makes both of these unreachable by default:
- `_spawn_finish`, the one place the child is started, runs in-process;
- `_rebuild_and_restart` raises unless a test patches it. The one group that exercises the real function (`rebuild` in `test_tool_updates.py`) gets it through `inspect.unwrap`, with every step faked.

With that in place, the hand-off mutation fails in 1.2 s and touches nothing.

## Release note

The v0.18.3 release is currently a **draft**, so updaters resolve to 0.18.2. It carries a recovery note (`quern setup`, then `quern restart`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update reliability by completing updates in a fresh process.
  * Added compatibility handling for updates initiated by older releases.
  * Update failures now provide clearer recovery guidance, including commands to complete setup and restart.
* **Documentation**
  * Added an Unreleased changelog entry documenting the update-crash fix.


jerimiah797:
@coderabbitai review

Issue #214, opened 2026-09-17
Setup accepts any Node version, but the MCP wrapper needs Node 22+


jerimiah797:
## What happened

A field machine with Node 20 (and Node 10 recorded before it) updated and ran setup. Setup reported:

```
✓ Node.js: v20.20.2
```

The MCP wrapper requires Node ≥ 22 (`mcp/package.json` `engines`, and `REQUIRED_MAJOR = 22` in `mcp/src/launcher.cjs`), so every MCP client launching it with that `node` gets the launcher's "requires Node >= 22" error, while setup says all is well. `npm install` only warned (`EBADENGINE`) and the build succeeded, so nothing else noticed either.

`check_node()` (`server/lifecycle/setup.py`) only asks whether `node` exists.

## Proposed

- **Setup:** below 22 is an error with instructions (install Node 22+, or point the MCP client's `command` at a 22+ binary), not a ✓. The "install via Homebrew?" offer should cover *too old* as well as *missing*.
- **Update (git installs):** check before pulling. The build after the pull needs npm and a supported Node, and stopping before the pull avoids a half-applied update.
- **Tarball installs:** installing needs no Node, since `mcp/dist` ships prebuilt. Running the MCP wrapper still does, so this is a setup error but not a reason to refuse the update.

## The caveat that shapes it

The check has to be about the `node` that will actually run, and that is not always the one on the caller's PATH. A menu-bar launch sees launchd's minimal PATH, and may find an old system Node while the user's fnm/nvm Node 22 is fine (#193). A hard stop there would block a working machine. So: halt when run from a terminal; from the menu bar, warn and say to run it from a terminal (`invoked_by()` already distinguishes the two). MCP clients also choose their own `node`, so name the binary that was checked.

Found alongside #212 (the update crash), which was the actual failure on that machine.

jerimiah797:
## Plan: diagnose all four places, warn instead of blocking

### There are four places a `node` gets picked, not one

Measured on the maintainer's machine after a Node cleanup:

| Place | Who uses it | What it finds | How to check |
|---|---|---|---|
| **The process running setup/update** | building `mcp/dist` (git installs) | whatever the caller had | `shutil.which` — the only check today |
| **Login shell** (`.zprofile` + `.zshrc`) | new Terminal windows, and CLI MCP clients started from them | fnm Node 22 | `$SHELL -lic 'command -v node; node --version'` |
[...]

PR #221, opened 2026-09-17
Update git installs in Terminal from the menu bar, and show the install type

jerimiah797:
Part of #214. This replaces PR 2 of the [plan](https://github.com/quern-dev/quern/issues/214#issuecomment-5716092345) (a dialog) with something smaller, decided with the maintainer: **on a git install, the menu bar doesn't try to update; it opens Terminal to do it.**

## Why the menu bar can't update a git install reliably

- **npm:** a git update rebuilds `mcp/dist` with npm, and GUI apps get launchd's PATH. On the maintainer's machine, Node 22 is found in every shell and **no `node` at all** by GUI apps (fnm lives in the shell's startup files).
- **Credentials:** `git pull` may need a credential prompt, and a menu-bar app has nowhere to show one.

A release install is fine from the menu: the tarball ships `mcp/dist` prebuilt.

## What changes

- **`InstallKind`** works out the install type from what the app already resolves.
  - The wrapper's `# Points to: <root>` line names the install; a release's `.venv/bin/quern-debug-server` implies the release folder.
  - A `.git` entry in that root means a git checkout. It's an existence check, because a worktree's `.git` is a file.
  - Anything unrecognised is *unknown*, and keeps today's behaviour.
- **Menu:** on a git install the staged-update item is **Update in Terminal… — vX** plus **Why Terminal? (git install)**, which links to the guide. Release and unknown installs keep **Restart to Update**. The decision is in `UpdateMenuItem.forStaged`, so it can be tested.
- **How Terminal gets the command:** a `quern-update.command` file in the temp folder, opened with Terminal. It runs the resolved wrapper by its quoted absolute path and says whether the update succeeded.
  - **Not AppleScript**, although that was the first plan: scripting Terminal from a hardened-runtime app needs the `apple-events` entitlement, an `NSAppleEventsUsageDescription`, and a permission prompt the user must accept; until then it fails with -1743. The app has none of those today. Opening a document needs none of them, and the result is the same Terminal window.
- **The window stays open.** Terminal's default profile closes a window whose shell exits cleanly, so the maintainer's successful test run vanished before it could be read. The script now ends with `exec "${SHELL:-/bin/zsh}" -l`: the output stays on screen and the window becomes an ordinary shell for follow-ups like `quern doctor`. A "press Return" pause would only have kept the output.
- **Settings:** a new **Install** row, e.g. `Git checkout — ~/Dev/quern` or `Release — ~/.local/share/quern`.
- **Docs link fixed:** the menu's **Documentation** item and the Settings button opened `https://quern.dev/docs`, which is a **404**. Both now open the installation guide.
- **Docs:**
  - the menu-bar guide gains **Updating**: why git installs use Terminal, and how to switch a git install to a release install (`quern stop` then the install one-liner) and back (`./quern setup` in the clone);
  - the getting-started guide no longer says the installer "clones" Quern; it downloads a release.

## Verification

- **Swift tests:** 103 passed, 10 new. Covered: a clone, a worktree, a release wrapper, the release venv binary, a release folder that is also a checkout, unresolved or unrecognised installs, the menu decision for each kind, script quoting (a path containing `'`), the script ending in a shell rather than exiting, and the Settings row.
- **App build:** `build.sh` builds the app.
- **Live, on this machine:**
  ```
  resolved: ~/.local/bin/quern
  install: git(~/Dev/quern) | display: Git checkout — ~/Dev/quern
  menu: updateInTerminal("Update in Terminal… — v0.18.5")
  open result: ok      → Terminal opened quern-update.command, which runs '…/.local/bin/quern' update
  ```
- **Live, switching an install:** in a sandbox with its own `HOME` (a wrapper pointing at a clone, stub `open`/`sudo`/…), I ran the real `install.sh` against 0.18.4 in a pty, answering yes only to creating the venv:
  - the wrapper was re-pointed to `~/.local/share/quern`;
  - setup completed and Claude Code was re-registered;
  - `Quern.app` was installed into `~/Applications` (only the launch was stubbed);
  - `quern version` → `0.18.4`.

  The pty wrapper reported exit 1 even though the installer printed its success message and ends in a `printf`. I didn't track that down; it's most likely the harness (`yes` through `script`).

**Not verified:** clicking the actual menu item in a running build. That needs the new app installed over the maintainer's running one.

## Summary by CodeRabbit

- **New Features**
  - Git installations now offer **Update in Terminal…** for staged updates, while release installations retain **Restart to Update**.
  - Settings identifies whether Quern is installed from Git or as a release.
  - Menu-bar and Settings documentation links now open the installation guide.
  - Added guidance for updating Git installations and switching to release installs.

- **Documentation**
  - Clarified that the installer downloads releases rather than cloning the repository.
  - Added menu-bar update instructions.


jerimiah797:
@coderabbitai review

PR #226, opened 2026-09-17
Offer a Terminal recovery button when the server does not come back

### LINKED ISSUE #225: Offer a Terminal recovery button when the server does not come back (especially after a menu-bar update)
## What happens today when the server doesn't come back after a menu-bar update

| Where it fails | What the user gets |
|---|---|
| During `quern update`: its own server restart fails | the alert **"The update did not complete"**, with the CLI output. No next step |
| After the update: the relaunched app's automatic start fails | **no alert**; the automatic start reports only in the menu, so a red icon, "Could not start the server", and **Open Server Log** |
| Any start the user clicks | an alert with the CLI output, and a red icon |
| `quern` not found | the menu line "quern not found — run `quern setup`" |

The worst of these is the second: the user has just clicked Update and is watching, but the app treats the failed start like an unattended one at login. None of them offers a way into recovery beyond reading a log.

## Proposal: a Terminal button everywhere, never an automatic Terminal

Decided with the maintainer: recovery is always **one click away**, and Terminal never opens unprompted (in particular not at login, where start-on-launch also runs).

1. **Menu:** whenever a start has failed, show **Troubleshoot in Terminal…** next to **Open Server Log**.
2. **Alerts:** every start, restart or update failure alert gets a **Fix in Terminal** button next to OK.
3. **After an update:** if the app launches within a few minutes of a recorded update (`~/.quern/last-update.json`, `outcome` + `finished_at`), a failed automatic start is reported with the alert, not only in the menu.

Terminal opens a `.command` script, reusing #221's `TerminalUpdate` mechanism (no Apple Events, and the window stays open as a shell afterwards). What it runs depends on what failed:

| Failure | Script |
|---|---|
| Update did not complete | `quern setup` then `quern restart`, the recovery for an update that stopped partway (#212) |
| Server would not start | `quern doctor --fix` (repairs a stale venv; after #223 also a missing or old menu-bar app) then `quern start` |
| `quern` not found | the setup command by absolute path, when an install root is known; otherwise the install one-liner |

Each script prints what it's about to do and why, runs the wrapper by its absolute path, and ends in an interactive shell with a pointer to `~/.quern/server.log`.

## Tests
- The choice of script for each failure.
- Which failures get an alert (the post-update window) and which only the menu (a login start outside that window).
- That the button is offered on every failure alert.
- That the menu item appears exactly when a start has failed.

jerimiah797:
Closes #225.

## The problem

When the server didn't come back, the menu bar offered a red icon and a log. The worst case: after a menu-bar update, the relaunched app's automatic start failed **with no alert at all**, because automatic starts only report in the menu, even though the user had just clicked Update and was watching.

## What changes

Decided with the maintainer: **a button everywhere, and Terminal never opens without a click.**

- **Menu:** after a failed start, **Troubleshoot in Terminal…** next to **Open Server Log**; **Set Up in Terminal…** when `quern` can't be found.
- **Alerts:** every failure alert that has a recovery offers **Fix in Terminal** straight after OK (before Copy and Open Log). The choice of buttons is in `FailureAlert.buttons`, so it can be tested. The old comment arguing against a Terminal button is replaced, because its reasons no longer apply: there's no Automation prompt, and the commands are fixed and run in a window the user watches.
- **After an update:** a failed automatic start within **5 minutes** of a recorded update (`last-update.json`, `updated` or `failed`, not `no_op`) shows an alert. Outside that window, e.g. at login, it stays menu-only.
- **What each recovery runs** (`Recovery.script`, wrapper by quoted absolute path, window left open as a shell, the server log path printed):

| Recovery | Offered when | Runs |
|---|---|---|
| `repair` | the server didn't start | `quern doctor --fix`, then `quern start` |
| `finishUpdate` | `quern update` didn't complete | `quern setup`, then `quern restart` (#212) |
| `setUp` | no `quern` found | the release install's own `…/quern setup` if it exists; otherwise **prints** the install one-liner rather than running it |

- **Shared Terminal helper:** `TerminalScript` is extracted from #221's update item (`TerminalUpdate` now uses it), so both write a `.command` file, open it with Terminal, and end in `exec "${SHELL:-/bin/zsh}" -l`.
- **`LifecycleController`:** now carries `recovery`, set alongside the failure it describes and cleared by a success or by the server appearing. `onAlert` passes it along.
- **Docs and changelog:** `docs/guides/menu-bar-app.md` and the CHANGELOG.

## Verification

- **Swift:** 116 passed, 13 new.
  - **Recovery scripts:** command order; the window stays open and the log is named; set-up with and without a release install; the installer shown and never run; the fallback to `quern` on PATH.
  - **Alert buttons:** the full set and the minimal set.
  - **Post-update window:** just updated, just failed, an old update, a no-op, no timestamp, a future timestamp, no record.
  - **Controller:** a failed start gives `repair`; a failed login start still offers it in the menu; a missing CLI gives `setUp`; a failed stop gives none; the recovery is cleared by the server appearing, and by a later success on the same controller.
- **Mutations, all 10 caught:** no recovery on a failed start; `repair` for a missing CLI; the alert dropping the recovery; a success keeping it; no Fix button; launch always quiet; a no-op counted as an update; start before doctor; the installer executed; the window closing. The "success keeps it" mutation initially survived because the test used a fresh controller; the test now fails and then succeeds on the same one.
- **Live:** `build.sh` builds the app. On this machine, `Recovery.repair.open` opened Terminal running the script shown in the diff, with the real wrapper path; the running server stayed healthy.
- **Not verified live:** clicking the buttons in a real failed-update scenario.

**Overlap:** this touches `AppDelegate.swift` and `docs/guides/menu-bar-app.md`, as #223 does; whichever merges second needs a rebase.

## Summary by CodeRabbit

- **New Features**
  - Added **Finish Update in Terminal…** for updates that stop partway, available whether or not the server is running.
  - Added terminal recovery options for start and update failures when recovery is available.
  - Added **Set Up in Terminal…** when the command-line tool is unavailable.
  - Terminal opens only after a recovery option is selected.
  - Recent update-related start failures show alerts; login-time failures remain menu-only.
  - Update recovery clears after the server returns on a different version.

- **Tests**
  - Added coverage for recovery actions, alerts, setup scenarios, and update failures.


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
Fixed in 123eb99:

- **Menu-bar guide overclaim** — now "Every start or update failure that has a way out offers the same step as a **Fix in Terminal** button."
- **CHANGELOG duplicate `### Added`** — folded into one section.
- **Update recovery cleared by `noteServerRunning()`** — good catch, and the drawing condition was the bigger half of it: an update that stops partway usually leaves the server *running*, so the item was inside `if !s.running` and never drawn in the common case. The two failures are now tracked apart (`recovery` vs `updateRecovery`), the update one is drawn outside that section and cleared when the user retries the update rather than when the daemon reports healthy. Since both items can be on the menu at once, the recovery now rides on the `NSMenuItem` instead of one shared `offeredRecovery` property, which the second writer was overwriting. Mutation-tested: restoring `updateRecovery = nil` in `noteServerRunning()` fails "a running server does not erase an update's way out".

On the expected test count: the count was correct at 124, not 125 — `grep -c "Harness.test("` across the test files totalled exactly 124 (19+3+10+10+17+10+34+21) and the suite reported "124 passed". This commit adds three cases, so it is now 127, and the suite reports 127 passed.

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
Both behaviour findings were real; fixed in 19a318d.

**Retire `.repair` when the server returns** — confirmed, and taken one step wider. `Recovery.forUpdateFailure` hands back `.repair` when the update failed *before* the restart, so it lands in `updateRecovery` where the version guard can never retire it — repairing a server does not change its version. Rather than special-casing `.repair`, the rule is now "only `.finishUpdate` survives a healthy server": anything else recorded there is about the server being down, and a server that is up has answered it. That covers `.setUp` on the same argument. Its baseline is cleared with it, so the next update failure is not measured against a version recorded for an unrelated one.

**No baseline when no version is cached** — confirmed, and the reachable case is the one you name: a git install with a `head_sha` and nothing that parses a version. The first reading after the failure now becomes the baseline. It cannot retire anything by itself (no change yet), and if a version never becomes readable the way out stays, which is the honest answer to "could not ask".

Mutation-tested both: `staleUpdateRecovery = false` fails "only an update's way out survives a healthy server"; dropping the baseline adoption fails "a version unknown at the failure is baselined, not guessed".

**Expected test count** — still declining this one. It was 124 when first raised, not 125; it is 141 now, and `grep -c "Harness.test("` across `macos/QuernMenuBar/Tests/*.swift` sums to exactly 141, which is what the suite reports. The count is derived from the source each time rather than guessed, so a bare number in a review is not evidence against it.

This push also carries fixes from a separate review pass: the update recovery had nothing to retire it on success at all, both recoveries rendered with the identical title "Troubleshoot in Terminal…" (the update's is now "Finish Update in Terminal…"), and the row decision moved into `FailureMenu.rows` so it is testable — three mutations of the previous commit, including moving the update row back inside `if !s.running`, had shipped green because nothing can construct an `AppDelegate`.

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

Issue #219, opened 2026-09-17
Rehearse every release's upgrade paths before tagging, including users' Node setups


jerimiah797:
## Why

Every release has turned up a problem with the *move* to the new version, and each was found by a user or by accident:

| Release | Problem | Why nothing caught it |
|---|---|---|
| 0.14.0 | `release/beta` not moved; beta users stranded | the steps are manual, in an order that's easy to get wrong |
| 0.18.0 → 0.18.1 | the beta channel offered a downgrade | nothing checked what each channel resolves to after publishing |
| 0.18.2 → 0.18.3 | the menu bar couldn't start a tarball install (npm not on the GUI PATH) | never started from a GUI-style environment |
| 0.18.3 (caught before tagging) | the tarball's MCP wrapper shipped without its dependencies | the check ran in a directory that happened to have them |
| 0.18.3 | **every update into it crashed** (#212) | nobody updated *from* the previous release |
| 0.18.3 | setup could leave the menu bar stopped (#215) | only seen during a manual test |
| field | Node 20 reported as fine; a GUI client with no Node at all (#214) | setup checked one place, and never the version |

The common thread: **we test the new code, not the move to it.** Updating, first installs, GUI launches and the user's Node setup were only ever tested by users.

### 1. Rehearsal before tagging: `scripts/release-rehearsal.sh <candidate>`

Automated, a few minutes. Each case runs in its own sandbox: its own `HOME` (with the dotfiles the case needs) and `QUERN_STATE_DIR`, stub `osascript`/`open`/`sudo`/`launchctl`/`pkill`/`killall` first on PATH, and a port that isn't 9100.

**Update and install**

| Case | Runs | Must hold |
|---|---|---|
| Git update | previous tag → candidate through a local bare repo, using the **old** updater | exit 0, no traceback, setup ran, `last-update.json` = `updated` with the new version |
| Tarball update | previous published tarball → candidate tarball served locally, using the **old** updater | same |
| Fresh install | `install.sh` → candidate tarball served locally | installs, setup runs, the wrapper works from another directory |
| GUI-style start | `quern start` under `env -i` with launchd's PATH | server healthy on the new version (#193) |
| Menu-bar step | setup with the app "running" and nothing new delivered | app not quit (#215) |

[...]

Issue #193, opened 2026-09-15
A tarball install cannot start from the menubar app when node came from fnm or nvm


jerimiah797:
Field report: a tarball install at v0.17.0 updated to v0.18.2. The update reported success, but the server never came back — the menubar app stayed grey. Starting it from the menubar turned the icon red and produced:

```
  File "…/server/main.py", line 767, in _cmd_start
    if not _ensure_mcp_built(quiet=True):
  File "…/server/__main__.py", line 232, in _ensure_mcp_built
    result = subprocess.run(
FileNotFoundError: [Errno 2] No such file or directory: 'npm'
```

`quern start` from a terminal on the same machine worked.

## Three things, compounding

**1. `_ensure_mcp_built` raises instead of returning False.** It shells out to npm twice and catches nothing. Every caller already treats a failed build as survivable:

```python
if not _ensure_mcp_built(quiet=True):
    print("Warning: MCP server build failed — MCP tools may be stale")
```

That intent only ever covered the failures the function *anticipated* — no project root, no `src/`, npm exiting non-zero. A **missing** npm raises, so the warning path is bypassed and the server dies.

This was already known. `server/lifecycle/updater.py:582` wraps the same call in `except (OSError, subprocess.SubprocessError)` with a comment describing this exact failure — *"a machine without npm … crashed `quern update`"*. The guard went at one of three call sites instead of into the function, so `quern update` survived and `quern start` did not. Same shape as the certificate gate: a guard on one path does not cover its siblings.

That also explains the sequence. The update itself survived because of that guard; its restart (`python -m server restart` → `_cmd_start`) hit the unguarded path, which is why the menubar stayed grey before anyone clicked anything.

**2. npm is unreachable from a GUI launch, and no PATH list can fix it.** The menubar app spawns the server with an augmented PATH, and the comment shows the problem was anticipated:

```swift
/// …so it can find python/git/etc. even when launched from a GUI context
/// (which has a minimal PATH).
let extra = ["\(home)/.local/bin", "/opt/homebrew/bin", "/usr/local/bin", "/usr/bin", "/bin"]
```

That finds a Homebrew node. It cannot find one installed by fnm or nvm — and neither can any other static list, because fnm hands out a directory named for the pid of the shell that asked. Quern already knows this: `tool_versions.py` classifies such paths as volatile, and `~/.quern/tool-sites.json` on an fnm machine records

```json
[...]

PR #229, opened 2026-09-18
Rehearse and verify a release, and pin where releases are fetched from

jerimiah797:
Layers 1 and 2 of #219, plus the product change they needed.

Every release so far has shipped a defect in the *move* to it rather than in the code: a channel branch left behind (0.14.0), a beta channel offering a downgrade (0.18.1), a tarball whose MCP wrapper had no dependencies (0.18.3), and an update that crashed for everyone (#212). Each was found by a user. This is the machinery that would have caught them.

## `QUERN_RELEASES_URL`

One variable, read by the updater, setup's app fetch, `quern menubar install` and the installer, so a rehearsal can serve a candidate locally. Unset, everything talks to GitHub exactly as before.

With it comes a trust check on the URLs a release response names, since what gets downloaded is extracted over the install. It is pinned to a prefix allowlist on `quern-dev/quern` rather than a host check — a host check alone accepts `https://github.com/someone-else/evil/...`, and a free account is enough to host a payload on a genuinely trusted host.

Review of that check found it bypassable, which is fixed here: `https://api.github.com/repos/quern-dev/quern/tarball/../../../octocat/Hello-World/tarball/master` passed it and fetched another repository's tree. api.github.com resolves dot segments server-side and urllib sends the path unchanged. Dot segments are now refused rather than normalised, and the compare uses the raw path so an encoded separator cannot read as a boundary here and as one character to the server.

## `scripts/release-rehearsal.sh` — before tagging

Runs the **previous release's** updater against the candidate, because that is the code every user runs. #212 was fine to install and crashed every update into it; no test of the new code could have seen that.

Proof it works: `scripts/release-rehearsal.sh v0.18.3 v0.18.2` still reproduces the field crash, `ImportError: cannot import name 'quern_cmd' from 'server.config'`, and fails four checks.

Cases: a git update, a GUI-style start (#193, `env -i` and launchd's PATH, asserting `/health` answers), the MCP wrapper answering `initialize`, the Node arrangements of #214 built to order in throwaway homes, and a fresh install through the site repo's `install.sh` against a locally served candidate. A tarball update is declared and skipped until the previous release honours the override.

Each case runs in its own sandbox — own `HOME` and `QUERN_STATE_DIR`, with `osascript`/`open`/`sudo`/`launchctl`/`pkill`/`killall` stubbed ahead of the real ones — and the run ends by checking nothing outside it was touched.

## `scripts/release-verify.sh` — after publishing

The API's latest release, what the released updater offers on both channels, the refs, the tarball's contents, the signed and notarized app, and quern.dev's answer to an old install.

Running it in rehearsal mode found it reporting passes it had not earned: the resolver checks run the *released* updater, which for anything before 0.18.5 asks GitHub, so against a locally served candidate it answered about the published release while printing green. Those now print as skips naming why, and skips are counted so a run with them cannot read like a clean one.

## Verified

- Full suite 2790 passed; Swift 141 passed.
- `release-verify.sh v0.18.4` passes against the real release; `v0.18.3` correctly fails six checks.
- `release-rehearsal.sh HEAD v0.18.2` passes all 26 checks.
- The traversal fix and three other guards are mutation-tested.

## Summary by CodeRabbit

- **New Features**
  - Release downloads can use a configured release service across updater, installation, and menu bar app flows.
  - Download links are validated before retrieval to help prevent downloads from untrusted locations.

- **Release Tools**
  - Added pre-release rehearsal checks for updates, fresh installations, startup, MCP initialization, and environment scenarios.
  - Added post-release verification checks for published versions, update channels, packaged assets, app signing, and website availability.

- **Documentation**
  - Updated release procedures with rehearsal and verification steps.


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
All three fixed in 2991acf.

- **The `install-menubar-app.sh` fallback** — good catch, and it is the sharper half of the finding: falling back to the github.com literal is right for an ordinary install from a clone with no venv, and wrong when `QUERN_RELEASES_URL` is set, because it installs the *published* app while the operator believes they are checking a candidate. That is the exact substitution the override exists to make visible. It now refuses in that case rather than falling back.
- **Snapshot filesystem metadata** — confirmed, and worse than it looks: `shasum` follows a symlink, so a link repointed at another tree reports on the wrong file without complaint. The snapshot records the link target and the mode now. I also added `~/.claude/skills/quern-api` to the watched set — a review agent repointed that exact symlink at a temporary directory earlier today by running a real setup by accident, so it is not a hypothetical shape.
- **Duplicate `### Added`** — folded.

The two you marked resolved were both found independently by a review agent and fixed in a832b73, along with the more serious one neither pass named: `quern start` reclaims its port by SIGKILLing whatever quern-looking process holds it, without consulting `QUERN_STATE_DIR`, so the sandbox server taking the default 9100 would have killed the developer's own server and its proxy. The sandbox now starts on 9187/9188, and the run asserts your server is still alive at the end. Verified just now on a run with a real server up — pid 21955 survived.

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

PR #230, opened 2026-09-18
Answer setup's prompts with -y, and stop asking about the venv

jerimiah797:
`quern setup` asks before several things, and with no terminal `_prompt_yn` declines rather than hanging. One of those prompts creates the virtualenv — so the `curl | bash` one-liner in a provisioning script or a Dockerfile installed a tree that could not run and said only "re-run `quern setup`". The release rehearsal (#219) caught it on a fresh install.

## The venv is not a question

It lives inside the install directory and every check past it needs it, the way `node_modules` is part of `npm install`, and there was never a second answer. It is created.

The branch that handled declining is gone, but the *stop* it existed for is kept: a venv that cannot be created ends the run there, rather than falling through to the block commented "we're inside the venv" and dying several hundred lines later on `ModuleNotFoundError: No module named 'httpx'`.

## `-y` covers the rest

Meaning what `apt-get -y` means: take each prompt's default. Not "yes to anything that does not need sudo" — installing a MITM certificate authority into booted simulators needs no sudo, and is exactly the decision that must not be made on someone's behalf.

Five prompts are marked `deliberate` and no flag answers them:

| Prompt | Why |
|---|---|
| the capture CA | a trust anchor that outlives the capture window |
| the tunneld LaunchDaemon | runs as root at boot, survives reboots, needs sudo |
| `sudo pipx install --global` | writes outside `$HOME`, needs sudo |
| the macOS crash dialog | user-wide, and `quern uninstall` never reverts it |
| `xcode-select --install` | opens a dialog somebody has to click |

Three of those need sudo or a GUI click that no flag can supply, so answering yes would buy a hung run where the old behaviour declined cleanly. So the rule is scope *and* privilege, not one instead of the other.

## Setup honours `auto_install_cert`

CONTRIBUTING says the setting means the same thing at every place that asks, and "a setting honoured in four of five places is worse than one honoured nowhere". Setup's prompt was the fifth and ignored it outright: a user who had turned it on was asked anyway, and one who had turned it off was asked again — which is how a considered no becomes a tired yes.

Reading it needs a third answer `get_auto_install_cert()` cannot give, since it folds unset into False (correctly — its callers ask "may I install without asking", and silence is not permission). `auto_install_cert_choice()` returns `True`, `False` or `None`. Setup reads it and never writes it: answering one prompt is not choosing a standing policy.

## The entry points parse their arguments

They dropped everything after the subcommand, so `quern setup --help` ran a full setup, `quern mcp-install --help` rewrote every MCP client config, `quern update --help` ran a real update, and a mistyped flag did the same silently — worse than refusing, because the caller believes they opted in. They share one checker now. `server/main.py` kept argparse's leftovers only for the no-subcommand case and discarded them otherwise.

## From review

- `-y` was dropped at the venv re-exec, which is where almost every prompt lives — so on a fresh install, the case it exists for, it reached one question out of a dozen.
- Two messages said the opposite of what happened: the banner claiming nothing could be asked, and a closing report blaming a missing terminal and prescribing `quern setup`, which under `-y` prints the same thing forever.
- Two mutations had survived: disconnecting the flag from `_prompt_yn` entirely, and removing it from either entry point. Every test patched the module variable rather than going through `run_setup`, and nothing exercised the argument parsing.

## Verified

2825 tests pass, ruff clean. Each fix is mutation-tested, including every `deliberate` marking — that test also requires its anchor to be unique, because the first version matched an earlier line in the file and passed against an unmarked prompt.

## Summary by CodeRabbit

- **New Features**
  - Added `quern setup -y`/`--yes` for unattended setup using default answers.
  - Setup now honors the saved certificate-install preference without changing it.
  - Setup creates the virtual environment without prompting and carries unattended mode across setup stages.

- **Bug Fixes**
  - Help requests, invalid options, and unexpected extra arguments are now handled correctly, preventing unintended command execution or ignored input.

- **Documentation**
  - Added README guidance for unattended setup and prompts that require explicit decisions.


jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
Both fixed in 7d083af.

- **Reject unexpected operands** — confirmed, and it is the same silent discard one level down from the one this PR set out to fix. `_check_args` refused unknown flags and then handed the leftover words to callers that dropped them, so `quern update typo` ran a real update. The three `set-*` commands were not going through it at all, and since each helper reads only its first argument, `quern set-channel stable typo` persisted `stable` while saying nothing about the word it did not understand. `_check_args` now takes an operand count: commands that take none say so, and the settings take exactly one and receive it. Tested both directions — the stray operand is refused, and the ordinary one-value call still reaches its helper.
- **Document every prompt `-y` leaves unanswered** — done, as a table. There are five, and they divide by two different reasons: the tunneld LaunchDaemon, the system-wide `sudo pipx` install and the Xcode CLT installer need a sudo password or a click that no flag can supply, so answering yes would leave the run waiting on something that never comes; the capture CA and the crash-dialog setting outlive quern itself.

2837 tests pass. (One full run showed two errors in `tests/test_device_controller.py::TestResolveUdid`; they pass in isolation and did not recur on a clean re-run, so that is a device-test flake rather than anything from this branch.)

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

jerimiah797:
@coderabbitai review

